PR #10, opened 2023-04-03
Improve Shared Library Loading Mechanism

MillionthOdin16:
This pull request refines the shared library loading code by introducing the `load_shared_library` function, which accepts the library's base name (`lib_base_name`) as input. Key improvements include:

1. Automatic determination of the library's file extension based on the operating system (`.so` for Linux, `.dylib` for macOS, `.dll` for Windows). This also solves an issue that occurs when more than one lib exists within the same dir.

2. Flexible library paths: Constructs two paths based on lib_base_name, accommodating slight differences in naming conventions used by `llamacpp` and `llama-cpp-python`.

3. On Windows (Python 3.8+), addition of the library directory to the DLL search path.

4. Attempted loading of the shared library via `ctypes.CDLL` and appropriate error handling, including raising `RuntimeError` for load failures and `FileNotFoundError` for missing libraries.

### DIFF
--- a/llama_cpp/llama_cpp.py
+++ b/llama_cpp/llama_cpp.py
@@ -1,17 +1,49 @@
+import sys
+import os
 import ctypes
-
 from ctypes import c_int, c_float, c_char_p, c_void_p, c_bool, POINTER, Structure, Array, c_uint8, c_size_t
-
 import pathlib
-from itertools import chain
 
 # Load the library
-# TODO: fragile, should fix
-_base_path = pathlib.Path(__file__).parent
-(_lib_path,) = chain(
-    _base_path.glob("*.so"), _base_path.glob("*.dylib"), _base_path.glob("*.dll")
-)
-_lib = ctypes.CDLL(str(_lib_path))
+def load_shared_library(lib_base_name):
+    # Determine the file extension based on the platform
+    if sys.platform.startswith("linux"):
+        lib_ext = ".so"
+    elif sys.platform == "darwin":
+        lib_ext = ".dylib"
+    elif sys.platform == "win32":
+        lib_ext = ".dll"
+    else:
+        raise RuntimeError("Unsupported platform")
+
+    # Construct the paths to the possible shared library names
+    _base_path = pathlib.Path(__file__).parent.resolve()
+    # Searching for the library in the current directory under the name "libllama" (default name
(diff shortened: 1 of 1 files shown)


abetlen:
Thank you for the contribution, I'll test this out shortly on a couple platforms and merge it in if everything works fine.
We have a few open issues around M1 support which I suspect are due to my hacky solution so hopefully this helps fix that.

llama_cpp/llama_cpp.py
```diff
@@ -1,17 +1,49 @@
...
-from itertools import chain
 
 # Load the library
-# TODO: fragile, should fix
-_base_path = pathlib.Path(__file__).parent
-(_lib_path,) = chain(
-    _base_path.glob("*.so"), _base_path.glob("*.dylib"), _base_path.glob("*.dll")
-)
-_lib = ctypes.CDLL(str(_lib_path))
+def load_shared_library(lib_base_name):
```
abetlen, line 8: Rename to `_load_shared_library` to avoid it appearing in any public APIs

llama_cpp/llama_cpp.py
```diff
@@ -1,17 +1,49 @@
...
+        if _lib_path.exists():
+            try:
+                return ctypes.CDLL(str(_lib_path))
+            except Exception as e:
+                raise RuntimeError(f"Failed to load shared library '{_lib_path}': {e}")
+
+    raise FileNotFoundError(f"Shared library with base name '{lib_base_name}' not found")
+
+# Specify the base name of the shared library to load
+lib_base_name = "llama"
```
abetlen, line 43: Same here, should be `_lib_base_name`

abetlen (changes requested):

abetlen:
Tested and works! Once those changes are in I'll merge this and push a new release to PyPI

llama_cpp/llama_cpp.py
abetlen: Sorry, I saw you marked this as resolved, is there a reason to not have this be `_load_shared_library`?

abetlen: @MillionthOdin16 oh I see the intention is to be able to patch the library with a custom dll at runtime correct?

abetlen: Nope, LGTM. Thank you for this!

Issue #14, opened 2023-04-03
[Question] How to use kv cache?


SagsMug:
Hello!

I have been trying to test the new kv cache loading and ran into an issue, it seems to segfault when running `llama_eval`.
To save the current cache i do: 
```python
import llama_cpp
import pickle
from ctypes import cast
# Some work...
kv_tokens = llama_cpp.llama_get_kv_cache_token_count(ctx)
kv_len = llama_cpp.llama_get_kv_cache_size(ctx)
kv_cache = llama_cpp.llama_get_kv_cache(ctx) 
kv_cache = cast(kv_cache, llama_cpp.POINTER(llama_cpp.c_uint8 * kv_len))
kv_cache = bytearray(kv_cache)
with open("test.bin", "wb") as f:
    pickle.dump([kv_cache,kv_tokens], f)
```
Loading:
```python
with open("test.bin", "rb") as f:
    kv_cache, kv_tokens = pickle.load(f)
    llama_cpp.llama_set_kv_cache(ctx, 
	    (llama_cpp.c_uint8 * len(kv_cache)).from_buffer(kv_cache),
	    len(kv_cache),
	    kv_tokens
    )
```
But running `llama_cpp.llama_eval` after will result in a segfault.

llama-cpp-python version: 0.1.16

How do i fix this?
Thanks

abetlen:
Unfortunately it looks like the `kv_state` API is still not enough to restore the model state [see issue](https://github.com/ggerganov/llama.cpp/issues/730#issuecomment-1493708919).

This is something I'm following so once it's possible I'll drop an example.

abetlen:
@SagsMug well this is emberrasing, so my c++ example from that thread apparently _was_ correct haha. Here is the general structure.

```
import time
import llama_cpp

llama = llama_cpp.Llama(model_path="../../models/ggml-alpaca.bin")
prompt = llama.tokenize(b"The quick brown fox")
# Reset the model
llama.reset()
# Feed the prompt
t1 = time.time()
llama.eval(prompt)
t2 = time.time()
### Save model state
assert llama.ctx is not None

# Save kv cache
kv_cache_token_count = llama_cpp.llama_get_kv_cache_token_count(llama.ctx)
kv_cache_size = llama_cpp.llama_get_kv_cache_size(llama.ctx)
kv_cache = llama_cpp.llama_get_kv_cache(llama.ctx)
[...]

Issue #7, opened 2023-04-01
Interactive mode/Session


BlackLotus:
Hey there sorry if this obviously possible/easy, but is it possible right now to use llama.cpp's interactive mode and if so how?

abetlen:
Hey, yes that's certainly possible. I don't have an example yet but this can be done through the [`generate`](https://abetlen.github.io/llama-cpp-python/#llama_cpp.llama.Llama.generate) method which returns a generator of tokens which you can also pass a new prompt too to continue generation.

BlackLotus:
Yeah was confused because generate only took the minimal vars. Thought about hacking an interactive mode into Llamas default __call__ itself something like this
https://github.com/BlackLotus/llama-cpp-python/compare/main...example
(DISCLAIMER: hacked together in a minute, untested, just for show, not usable code)
but since this seemed way too hackish I implemented a minimal PoC that does it's own "session management" like stop words.
Anyway thanks for the reply it really helped.

SagsMug:
I have tried to port the original main from llama.cpp to python here: https://gist.github.com/SagsMug/c39d5f337b0e0faf32b3ff6bc5bf4956
It seems to work okay

BlackLotus:
Is a session mode (calling it session here an interactive mode where since user interaction like ctrl+c is required probably isn't wanted in a python lib) wanted in this library.
Implementing a continuous mode through the use of a generator is probably a good enough solution the only "problem" I have with it is that I implemented my own parsing/interacting function, but hacking together an interactive mode should be relatively straight forward.... if it is wanted

BlackLotus:
> I have tried to port the original main from llama.cpp to python here: https://gist.github.com/SagsMug/c39d5f337b0e0faf32b3ff6bc5bf4956 It seems to work okay

Could be an example of using it as chat. :) thx for posting similar to my personal hacked together version, but I didn't bother to use hinting. Maybe I will use your "wrapper".

abetlen:
> I have tried to port the original main from llama.cpp to python here: https://gist.github.com/SagsMug/c39d5f337b0e0faf32b3ff6bc5bf4956 It seems to work okay

This is cool, I'd like to include python ports of the `llama.cpp` examples to give people an idea of how to use the low-level api. Gladly accept any PRs for this as well.

[...]

PR #15, opened 2023-04-03
llama.cpp chat example implementation

SagsMug:
This commit adds a port of llama.cpp's main function as an example.
It's finally reached a stage where its readable enough for general usage and learning.
There are some differences from the original main since i wanted programmatic I/O.

Future work:
- Implement a circular buffer
- Context saving for chat resuming

On the first, like the original we just use a list and pop the first element.
Python's deque doesn't support slicing and implementing a custom class seemed out of scope for an example.
It's not the slowest part anyway, since we're waiting for llama most of the time, so its not a high priority.

We can say that that's left as an exercise for the reader 😋

On the second, see https://github.com/abetlen/llama-cpp-python/issues/14

Also solves https://github.com/abetlen/llama-cpp-python/issues/7, unless we want a higher level interactive mode also


MillionthOdin16:
> Context saving for chat resuming

Awesome! I have a question about the `context` since you mentioned it. Using the `llama.cpp` main, I can have a continuous conversation with the model where the `context` is stored for up to `2048` tokens. The only time I lose the model's memory is when I terminate the executable. When I start the model again, it loads the prompt, and I start fresh like I did last time I started the memory.

Does your implementation of main still maintain context between individual messages? I ask because I was running into issues while testing. One of the big concerns I had while testing today was that I couldn't figure out why it wasn't maintaining context, and I definitely don't want to resend the full prompt and message history each time.

Other than that, I need to check in with llama.cpp to see where they are on state saving. I know lots of people are interested. Would def be a cool ability!

> Instruction mode

I do thing it would be a good idea to implement instruction mode and before people get the example. I think it's one of the more heavily use flags between `-i` and `-ins'. Is there something the makes it difficult?

Thanks :)

SagsMug:
> Does your implementation of main still maintain context between individual messages? I ask because I was running into issues while testing. One of the big concerns I had while testing today was that I couldn't figure out why it wasn't maintaining context, and I definitely don't want to resend the full prompt and message history each time.

This is the ```infinite text generation``` part of llama.cpp, and yes its implemented but untested, see: https://github.com/abetlen/llama-cpp-python/blob/ad54bad79262a723e88fbffd1ed338e873e5105d/examples/low_level_api_chatllama_cpp.py#L110-L123

> Other than that, I need to check in with llama.cpp to see where they are on state saving. I know lots of people are interested. Would def be a cool ability!

Definitely, that's also why i made the issue cause i thought that i could have state saving, but alas, it wasn't meant to be.

> I do thing it would be a good idea to implement instruction mode and before people get the example. I think it's one of the more heavily use flags between `-i` and `-ins'. Is there something the makes it difficult?

Its now implemented using ```instruct=True``` flag, there aren't any command-line flags right now.
One note is that you should feed the initial prompt with some text that shows it how to return to instruction mode, but you probably do that anyway for normal llama.
Otherwise it will infinitely generate nonsense, and you will have to implement a KeyboardInterrupt catcher to stop generation, and i will be sad.
(okay i implemented a KeyboardInterrupt catcher).

abetlen:
Good progress so far, I would like to include this in the examples I just request that we follow the `llama.cpp` examples more closely, specifically, could you instead store / load parameters into using something like the below:

```python
import os
import argparse

from dataclasses import dataclass, field
from typing import List, Optional

# Based on https://github.com/ggerganov/llama.cpp/blob/master/examples/common.cpp


@dataclass
class GptParams:
    seed: int = -1
    n_threads: int = min(4, os.cpu_count() or 1)
    n_predict: int = 128
    repeat_last_n: int = 64
    n_parts: int = -1
    n_ctx: int = 512
    n_batch: int = 8
    n_keep: int = 0

    top_k: int = 40
    top_p: float = 0.95
    temp: float = 0.80
    repeat_penalty: float = 1.10

    model: str = "models/lamma-7B/ggml-model.bin"
    prompt: str = ""
    input_prefix: str = ""

    antiprompt: List[str] = field(default_factory=list)

    memory_f16: bool = True
    random_prompt: bool = False
    use_color: bool = False
    interactive: bool = False

    embedding: bool = False
    interactive_start: bool = False

    instruct: bool = False
    ignore_eos: bool = False
    perplexity: bool = False
    use_mlock: bool = False
    mem_test: bool = False
    verbose_prompt: bool = False


def gpt_params_parse(argv, params: Optional[GptParams] = None):
    if params is None:
        params = GptParams()

    parser = argparse.ArgumentParser()
    parser.add_argument("-s", "--seed", type=int, default=-1, help="")
    parser.add_argument("-t", "--threads", type=int, default=1, help="")
    parser.add_argument("-p", "--prompt", type=str, default="", help="")
    parser.add_argument("-f", "--file", type=str, default=None, help="")
    parser.add_argument("-c", "--context_size", type=int, default=512, help="")
    parser.add_argument("--memory_f32", action="store_true", help="")
    parser.add_argument("--top_p", type=float, default=0.9, help="")
    parser.add_argument("--temp", type=float, default=1.0, help="")
    parser.add_argument("--repeat_last_n", type=int, default=64, help="")
    parser.add_argument("--repeat_penalty", type=float, default=1.0, help="")
    parser.add_argument("-b", "--batch_size", type=int, default=8, help="")
    parser.add_argument("-m", "--model", type=str, help="")
    parser.add_argument(
        "-i", "--interactive", action="store_true", help="run in interactive mode"
    )
    parser.add_argument("--embedding", action="store_true", help="")
    parser.add_argument("--interactive-start", action="store_true", help="")
    parser.add_argument(
        "--interactive-first",
        action="store_true",
        help="run in interactive mode and wait for input right away",
    )
    parser.add_argument(
        "-ins",
        "--instruct",
        action="store_true",
        help="run in instruction mode (use with Alpaca models)",
    )
    parser.add_argument(
        "--color",
        action="store_true",
        help="colorise output to distinguish prompt and user input from generations",
    )
    parser.add_argument("--mlock", action="store_true")
    parser.add_argument("--mtest", action="store_true")
    parser.add_argument(
        "-r",
        "--reverse-prompt",
        type=str,
        default="",
        help="run in interactive mode and poll user input upon seeing PROMPT (can be\nspecified more than once for multiple prompts).",
    )
    parser.add_argument("--perplexity", action="store_true", help="")
    parser.add_argument("--ignore-eos", action="store_true", help="")
    parser.add_argument("--n_parts", type=int, default=-1, help="")
    parser.add_argument("--random-prompt", action="store_true", help="")
    parser.add_argument("--in-prefix", type=str, default="", help="")
    args = parser.parse_args(argv)
    return args
```

Ideally place this into a `examples/low_level_api/common.py` file to follow how `main.cpp` works.

examples/common.py
```diff
@@ -0,0 +1,135 @@
+import os
```
abetlen, line 1: Can you move this common.py file to just the low_level_api folder and remove the __init__.py files?

abetlen (changes requested):
Can you move this common.py file to just the low_level_api folder and remove the init.py files?

examples/common.py
SagsMug: I moved common.py and removed the init.py files

SagsMug:
> Can you move this common.py file to just the low_level_api folder and remove the init.py files?

Done!
It should now very closely match the original implementation with arguments and everything.
Tested on Miku.sh from the original repo

abetlen:
@SagsMug thank you!

Issue #12, opened 2023-04-03
[Investigate] Custom `llama.dll` Dependency Resolution Issues on Windows


MillionthOdin16:
This is a note for using a custom `llama.dll` build on Windows. I ran into dependency resolution issues with loading my own `llama.dll` compiled with BLAS support and some extra hardware specific optimization flags. No matter what I do, it can't seem to locate all of its dependencies, even though I've tried placing them in system paths and even same dir.

My current workaround is using the default `llama.dll` that llama-cpp-python builds, but it doesn't have the hardware optimizations and BLAS compatibility that I enabled in my custom build. So, I'm still trying to figure out what my issue is. Maybe something python specific that i'm missing...

I'm dropping this issue here just in case anyone else runs into something similar. If you have any ideas or workarounds, let me know. I'll keep trying to figure it out until I get it resolved haha :)

geocine:
For me I could not even get past this
```python
# TODO: fragile, should fix
 _base_path = pathlib.Path(__file__).parent
(_lib_path,) = chain(
     _base_path.glob("*.so"), _base_path.glob("*.dylib"), _base_path.glob("*.dll")
 )](url)
```
I get this error
```sh
(_lib_path,) = chain(
ValueError: not enough values to unpack (expected 1, got 0)
```
So I just commented it out and did this instead
```py
_lib_path = "C:\path\to\my\llama.dll"
```

abetlen:
@geocine I've merged in @MillionthOdin16 fix for library loading, does the default installation path work now for you?

geocine:
Thanks @abetlen , I will try again later

MillionthOdin16:
I think there might be a combination of 2 things that make the lib setup confusing at the moment:

1. I don't know at which point llama-cpp-python compiles the `llama` lib, and when I first got setup I think I had to go and compile it after searching. So in this case, it might not actually exist.
2. When bringing a custom built 'llama' lib for extra functionality (like `BLAS`), it detects the `llama` lib itself, but can't find other dependencies like it normally should. I'm still trying to figure out if this is a Windows only type of thing or related to python somehow.

*Just a note, the `CMakeLists` for `llama.cpp` are less developed on Windows side than the MacOS/Linux side (which the devs are more focused on), so for Windows they often require manual tweaks to make them work correctly. This is one reason my first inclination was to bring my own lib. Just wanna make sure people know in case Windows users have more trouble.

abetlen:
[...]

PR #28, opened 2023-04-05
Allow local llama library usage

SagsMug:
Sometimes its nice to try different library versions and different forks of llama etc (given they still use the same api).
This commit adds the local path as the first place to search for llama libraries.

### DIFF
--- a/llama_cpp/llama_cpp.py
+++ b/llama_cpp/llama_cpp.py
@@ -18,9 +18,12 @@ def _load_shared_library(lib_base_name):
 
     # Construct the paths to the possible shared library names
     _base_path = pathlib.Path(__file__).parent.resolve()
+    _local_path = pathlib.Path.cwd()
     # Searching for the library in the current directory under the name "libllama" (default name
     # for llamacpp) and "llama" (default name for this repo)
     _lib_paths = [
+        _local_path / f"./lib{lib_base_name}{lib_ext}",
+        _local_path / f"./{lib_base_name}{lib_ext}",
         _base_path / f"lib{lib_base_name}{lib_ext}",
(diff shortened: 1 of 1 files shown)


abetlen:
I agree we may need an escape hatch of sorts at the moment (e.g #30 ) and @MillionthOdin16 has brought this up previously in #12 .

I'm not sure the that using the local directory is the right solution, I would prefer something more explicit such as using an environment variable or even a custom function to update `_lib` in `llama_cpp`.

SagsMug:
> I agree we may need an escape hatch of sorts at the moment (e.g #30 ) and @MillionthOdin16 has brought this up previously in #12 .
> 
> I'm not sure the that using the local directory is the right solution, I would prefer something more explicit such as using an environment variable or even a custom function to update `_lib` in `llama_cpp`.

I made it into an environment variable.
A custom function will probably take a bigger refactor because of the way the library is loaded.

llama_cpp/llama_cpp.py
```diff
@@ -25,6 +25,12 @@ def _load_shared_library(lib_base_name):
         _base_path / f"{lib_base_name}{lib_ext}"
     ]
 
+    if ("LLAMA_LIB" in os.environ):
```
abetlen, line 28: @SagsMug great work, can we change the environment variable to `LLAMA_CPP_LIB`

abetlen (changes requested):
@SagsMug great work

llama_cpp/llama_cpp.py
SagsMug: > @SagsMug great work, can we change the environment variable to `LLAMA_CPP_LIB`

Done!

PR #42, opened 2023-04-07
Use https for the gitmodule instead of ssh

nsarrazin:
## Why
The submodule is defined using SSH, which means that you cannot initialize the submodule if you don't have SSH set up. 

I was building `llama-cpp-python` from source inside of a docker container which doesn't have SSH set up and I couldn't clone the submodule.

## How
I just replaced the url in `.gitmodules`, it should have no other side effects!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "vendor/llama.cpp"]
 	path = vendor/llama.cpp
-	url = git@github.com:ggerganov/llama.cpp.git
+	url = https://github.com/ggerganov/llama.cpp.git


abetlen:
I probably won't merge this because I think git ssh URLs are the recommended approach. A few thoughts though 
- if you're building docker images there's a feature to forward ssh agents into the container
- pip installing the package already builds from source by default
- if you do get this working as a docker image without this change I'll definitely merge it in and add it to the build process

nsarrazin:
I think github doesn't specifically recommend ssh over https, [it even used to be the other way](https://stackoverflow.com/questions/11041729/why-does-github-recommend-https-over-ssh) :laughing: 


But for my project I don't specifically need to build the bindings from source, I just want to make sure `llama.cpp` gets compiled on the user's machine and is not shipping with a binary because there are [issues with CPU feature sets](https://github.com/ggerganov/llama.cpp/discussions/535) if you don't.

So if I do pip install it will compile llama.cpp from source ?

abetlen:
@nsarrazin yes that's correct, and I plan to keep that the default (no wheels in the PyPI package)

nsarrazin:
Alright that's great to hear. This will simplify my dockerfile, thanks haha

I'm closing this PR then!

Issue #40, opened 2023-04-06
Installation on Windows failed because Visual Studio is not installed


oobabooga:
Trying to install with

```
pip install llama-cpp-python==0.1.23
```

on Windows in a micromamba environment resulted in the following error. It seems like the package is looking for Visual Studio, which is not installed on my system.

Is it possible to make it such that the package can be installed without the need for Visual Studio?

```
(C:\Users\me\Downloads\oobabooga-windows\oobabooga-windows\installer_files\env) C:\Users\me\Downloads\oobabooga-windows\oobabooga-windows\text-generation-webui>pip install llama-cpp-python==0.1.23
Collecting llama-cpp-python==0.1.23
  Downloading llama_cpp_python-0.1.23.tar.gz (530 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 530.0/530.0 kB 504.2 kB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: typing-extensions>=4.5.0 in c:\users\me\downloads\oobabooga-windows\oobabooga-windows\installer_files\env\lib\site-packages (from llama-cpp-python==0.1.23) (4.5.0)
Building wheels for collected packages: llama-cpp-python
  Building wheel for llama-cpp-python (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for llama-cpp-python (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [308 lines of output]


      --------------------------------------------------------------------------------
      -- Trying 'Ninja (Visual Studio 17 2022 x64 v143)' generator
      --------------------------------
      ---------------------------
      ----------------------
      -----------------
      ------------
      -------
      --
      Not searching for unused variables given on the command line.
      -- The C compiler identification is unknown
      CMake Error at CMakeLists.txt:3 (ENABLE_LANGUAGE):
        No CMAKE_C_COMPILER could be found.

        Tell CMake where to find the compiler by setting either the environment
        variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
        the compiler, or to the compiler name if it is in the PATH.


      -- Configuring incomplete, errors occurred!
      --
      -------
      ------------
      -----------------
      ----------------------
      ---------------------------
      --------------------------------
[...]

PR #50, opened 2023-04-09
Add abetlen's wheel to new /Wheels/ dir (removed from git ignore) + add amd64 c310 wheel

Loufe:
Following @abetlen's great work on https://github.com/abetlen/llama-cpp-python/issues/40 I did some testing and was able to modify Ooba's one-click install to use it. Works great on my branch.  

SHA256 of .whl: FEAAF0D4002018A15992B12CF57319CDCBAC27DE0C78CF86EB16EC6A58BD86A7


abetlen:
hey @Loufe can you check the release wheels to make sure I've added the correct ones, sorry there's so many, I went with the approach here https://github.com/pypa/cibuildwheel

Loufe:
@abetlen tested and working, LGTM! Good choice with Cibuildwheel.

I would save Github some power and leave it cleaner by limiting the job to Python 3.10. According to Pypa there is a param to limit version "[CIBW_BUILD](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip)", i.e.:
`CIBW_BUILD: cp310-*`

That said, I would even go so far as to limit the build to amd64 until suhc a time as someone requests a wheel for other OS / Windows archs. I.e.:
`CIBW_BUILD: cp310-cp310-win_amd64`

By the way, I wrote this call to dynamically find the most recent version considering the current tag name - if I merge this into Ooba's project should this be ok? I guess more specifically, you don't plan on changing build formats at any points?

abetlen:
@Loufe thanks! I'll leave the other wheels in case anyone needs them but I'll investigate which ones are completely redundent in the future.

PR #64, opened 2023-04-10
Add unlimited max_tokens

jm12138:
* Add a feature with unlimited max_tokens that is enabled when max_tokens <= 0. 

* Similar to this parameter in llama.cpp: 

https://github.com/ggerganov/llama.cpp/blob/180b693a47b6b825288ef9f2c39d24b6eea4eea6/examples/common.cpp#L230


Aqirito:
is this will be merged?

gjmulder:
@abetlen ?

BakaKiller:
Up ! This would be really great !

abetlen:
@jm12138 thanks for the contribution, sorry took so long to get to this! LGTM

PR #73, opened 2023-04-12
Add Dockerfile + build workflow

### LINKED ISSUE #70: Problems when i try to use this inside the default python 3.10 docker container
When i try to install and use this package via a requirements file in the default 3.10 python container i get the following error when i try to import the module:
`Failed to load shared library '/usr/local/lib/python3.10/site-packages/llama_cpp/libllama.so': /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version 'GLIBCXX_3.4.29' not found (required by /usr/local/lib/python3.10/site-packages/llama_cpp/libllama.so)`

Am i doing something wrong? Or am i just missing some dependencies?

Niek:
Fixes #70

This PR adds a Dockerfile and updates the release workflow to build the latest Docker image too. Both amd64 and arm64 arches are built.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM python:3-buster
 ENV HOST 0.0.0.0
 
 # Install the package
-RUN pip install llama-cpp-python[server]
+RUN apt update && apt install -y libopenblas-dev && LLAMA_OPENBLAS=1 pip install llama-cpp-python[server]
 
 # Run the server
 CMD python3 -m llama_cpp.server
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:3-buster
+
+# We need to set the host to 0.0.0.0 to allow outside access
+ENV HOST 0.0.0.0
+
+# Install the package
+RUN pip install llama-cpp-python[server]
+
+# Run the server
+CMD python3 -m llama_cpp.server
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3-buster
+FROM python:3-bullseye
 
 # We need to set the host to 0.0.0.0 to allow outside access
 ENV HOST 0.0.0.0
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -29,3 +29,32 @@ jobs:
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         password: ${{ secrets.PYPI_API_TOKEN }}
+
+  docker:
+    name: Build and push Docker image
+    runs-on: ubuntu-latest
+    needs: build-n-publish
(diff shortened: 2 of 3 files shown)


abetlen:
@Niek do you mind moving this to the build release workflow?

jmtatsch:
Maybe we should directly add openblas support?
would need those two lines:
```
RUN apt update && apt install -y libopenblas-dev
RUN LLAMA_OPENBLAS=1 pip install llama-cpp-python[server]
```

Niek:
Good idea @jmtatsch - added now

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3-buster
```
jmtatsch, line 1: why use buster? bullseye is the current Debian distribution.
why use 3 and not something more up to date like [3.10.11-bullseye](https://github.com/docker-library/python/blob/4c86fbef4c546a65e372c9bc8c27b8c17bfe42fd/3.10/bullseye/Dockerfile)

jmtatsch:
Here is a docker file for a cublas capable container that should bring huge speed ups for cuda gpu owners after the next sync with upstream:

```
FROM nvidia/cuda:12.1.0-devel-ubuntu22.04

EXPOSE 8000
ENV MODEL=/models/ggml-vicuna-13b-1.1-q4_0.bin
# allow non local connections to api
ENV HOST=0.0.0.0

RUN apt update && apt install -y python3 python3-pip

COPY requirements.txt requirements.txt
RUN LLAMA_CUBLAS=1 pip install -r requirements.txt && rm -rf requirements.txt

ENTRYPOINT [ "python3", "-m", "llama_cpp.server" ]
```

gjmulder:
Where is `requirements.txt `coming from?

gjmulder:
> Here is a docker file for a cublas capable container that should bring huge speed ups for cuda gpu owners after the next sync with upstream:

@jmtatsch where is `requirements.txt` coming from?

jmtatsch:
> @jmtatsch where is `requirements.txt` coming from?

good catch, it isn't necessary at all. I cleaned it up above.

Dockerfile
Niek: Good point! Fixed now

Niek:
@abetlen do you need any other changes?

abetlen:
@Niek if possible can we include @jmtatsch nvidia-docker container example as well in this PR? Ability to docker pull and run a GPU-accelerated container would be very helpful.

jmtatsch:
@abetlen We should make this two different containers then because the nvidia container with cublas is quite fat and not everyone has a Nvidia card.
I will make a pull request once this one is merged.
Sorry for hijacking your pull request @Niek

abetlen:
@Niek finally got a chance to merge this, great work! We now have a docker image.

PR #111, opened 2023-04-25
Support defaulting to infinity or -1 for completions

swg:
Hello,

As per llama.cpp help:

` -n N, --n_predict N   number of tokens to predict (default: 128, -1 = infinity)`

The OpenAI docs state that the /completions endpoint defaults to 16, but is optional:
`max_tokens
integer
Optional
Defaults to 16`
https://platform.openai.com/docs/api-reference/completions/create#completions/create-max_tokens

However, /chat/completions defaults to infinity, and is again optional:
`max_tokens
integer
Optional
Defaults to inf`
https://platform.openai.com/docs/api-reference/chat/create#chat/create-max_tokens

This pull request modifies the fast_api example, the main server and supporting python lib to allow passing a 'null' which will default to infinity, mimicking the OpenAI API.


examples/high_level_api/fastapi_server.py
```diff
@@ -156,7 +156,7 @@ class CreateChatCompletionRequest(BaseModel):
     top_p: float = 0.95
     stream: bool = False
     stop: List[str] = []
-    max_tokens: int = 128
+    max_tokens: Optional[Union[int, None]] = -1
```
peturparkur, line 159: The type signature for `max_tokens: Optional[Union[int, None]]` doesn't make any sense.
`Optional[T]` is equivalent/alias of `Union[T, None]`

K-Mistele:
Hey guys, seems like there's a related error in #983 where if you don't specify the `max_tokens` and then send a prompt larger than `n_ctx` you run into a `TypeError` due to server attempting to add the default value of `max_tokens` which is `None` to the number of tokens from the prompt.

abetlen:
@K-Mistele yup, I added a few commits to hopefully fix that in the server error handler.

@swg thanks for the contribution! Sorry it took so long to merge.

K-Mistele:
With these fixes what is now the proper way to set an unlimited number of response tokens?

abetlen:
@K-Mistele passing None / null for `max_tokens` should work everywhere now and is recommended as it matches the OpenAI spec.

Issue #36, opened 2023-04-06
Fix unicode decoding error


abetlen:
Certain tokens in the vocabulary cannot be decoded to valid utf-8, I'm actually not sure if this is because they represent partial utf codepoints, but in any case they cause generation to fail.

wataruY:
Probably due to '🦙' in README.md

abetlen:
Should be fixed with #118

Issue #116, opened 2023-04-26
UnicodeDecodeError: 'utf-8' codec can't decode byte 


mozzipa:
For trial to have streaming of response, below error occurs

```
File [~/miniconda/envs/fastapi/lib/python3.10/site-packages/llama_cpp/llama.py:482](https://file+.vscode-resource.vscode-cdn.net/Users/jinwhan/Documents/Coding/Solidity/Page/cloudRun/cloudrun-fastapi/app/~/miniconda/envs/fastapi/lib/python3.10/site-packages/llama_cpp/llama.py:482), in Llama._create_completion(self, prompt, suffix, max_tokens, temperature, top_p, logprobs, echo, stop, repeat_penalty, top_k, stream)
    473     self._completion_bytes.append(text[start:])
    474     ###
    475     yield {
    476         "id": completion_id,
    477         "object": "text_completion",
    478         "created": created,
    479         "model": self.model_path,
    480         "choices": [
    481             {
--> 482                 "text": text[start:].decode("utf-8"),
    483                 "index": 0,
...
    488     }
    490 if len(completion_tokens) >= max_tokens:
    491     text = self.detokenize(completion_tokens)

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xec in position 0: unexpected end of data
```

My code snippet was prepared as below referring [Example](https://github.com/abetlen/llama-cpp-python/blob/main/examples/high_level_api/high_level_api_streaming.py).
```
from llama_cpp import Llama
import json


model_path = /my/model/path/for/ko_vicuna_7b/ggml-model-q4_0.bin"
prompt = "Tell me about Korea in english"
llm = Llama(model_path=model_path, n_ctx=4096, seed=0)

stream = llm(
    f"Q: {prompt} \nA: ", 
    max_tokens=512, 
    stop=["Q:", "\n"], 
    stream=True,
    temperature=0.1,
    )

for output in stream:
    print(output['choices'][0]["text"], end='')

```

Not only 0xec, but also 0xed, 0xf0 occurred for other trial cases. I cannot assure but it may be caused by language of [model](https://huggingface.co/Mozzipa/ko_vicuna_7b_ggml_q4) which is [fine tuned for korean](https://huggingface.co/junelee/ko_vicuna_7b) from vicuna 7b.

For your reference, several letters were given but it stops suddenly with above error.

gjmulder:
May be fixed in #118

Issue #100, opened 2023-04-21
Compatibility issues with Chinese and slow response speed.


B1gM8c:
There seem to be some compatibility issues with [ChatBot](https://github.com/mckaywrigley/chatbot-ui), as an error occurs when inputting Chinese characters.
```
    result = await app(  # type: ignore[func-returns-value]
  File "/usr/local/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/fastapi/applications.py", line 276, in __call__
    await super().__call__(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/cors.py", line 84, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/usr/local/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/usr/local/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/usr/local/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 718, in __call__
    await route.handle(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 69, in app
    await response(scope, receive, send)
  File "/usr/local/lib/python3.10/site-packages/sse_starlette/sse.py", line 227, in __call__
    async with anyio.create_task_group() as task_group:
[...]

Issue #57, opened 2023-04-09
Issue with emoji decoding in steaming mode, only


CyberTimon:
When the model wants to output an emoji, this error comes up: 

`Debugging middleware caught exception in streamed response at a point where response headers were already sent.
Traceback (most recent call last):
  File "C:\Users\zblac\AppData\Local\Programs\Python\Python310\lib\site-packages\werkzeug\wsgi.py", line 500, in __next__
    return self._next()
  File "C:\Users\zblac\AppData\Local\Programs\Python\Python310\lib\site-packages\werkzeug\wrappers\response.py", line 50, in _iter_encoded
    for item in iterable:
  File "C:\Users\zblac\llama.cpp\test\normal.py", line 37, in vicuna
    for line in response:
  File "C:\Users\zblac\AppData\Local\Programs\Python\Python310\lib\site-packages\llama_cpp\llama.py", line 370, in _create_completion
    "text": text[start:].decode("utf-8"),
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf0 in position 0: unexpected end of data`

Niek:
I can confirm this issue

PR #118, opened 2023-04-26
Fix UnicodeDecodeError permanently

SagsMug:
https://docs.python.org/3/library/codecs.html#error-handlers

This detects a multibyte UTF8 character, and doesn't return if its incomplete.
If there otherwise wasn't enough tokens or the model somehow returns invalid bytes, we use errors="ignore" to remove invalid characters.
Ascii example:
> 'German ß, ♬'.encode(encoding='ascii', errors="ignore")
> b'German , '

But there will never be thrown a decode or encode error.

Fixes:
https://github.com/abetlen/llama-cpp-python/issues/36
https://github.com/abetlen/llama-cpp-python/issues/57
https://github.com/abetlen/llama-cpp-python/issues/100
https://github.com/abetlen/llama-cpp-python/issues/116

### DIFF
--- a/examples/low_level_api/low_level_api_chat_cpp.py
+++ b/examples/low_level_api/low_level_api_chat_cpp.py
@@ -96,7 +96,7 @@ specified) expect poor results""", file=sys.stderr)
 
 		print(file=sys.stderr)
 		print(f"system_info: n_threads = {self.params.n_threads} / {cpu_count()} \
-| {llama_cpp.llama_print_system_info().decode('utf8', errors='ignore')}", file=sys.stderr)
+| {llama_cpp.llama_print_system_info().decode('utf8')}", file=sys.stderr)
 
 		# determine the required inference memory per token:
 		if (self.params.mem_test):
--- a/examples/low_level_api/low_level_api_chat_cpp.py
+++ b/examples/low_level_api/low_level_api_chat_cpp.py
@@ -356,10 +356,7 @@ n_keep = {self.params.n_keep}
 	def output(self):
 		self.remaining_tokens = self.params.n_predict
 		for id in self.generate():
-			try:
-				yield llama_cpp.llama_token_to_str(self.ctx, id).decode("utf-8", errors="ignore")
-			except UnicodeDecodeError:
-				pass
+			yield llama_cpp.llama_token_to_str(self.ctx, id).decode("utf-8", errors="ignore")
 
 	# read user input
 	def read_input(self):
--- a/examples/low_level_api/low_level_api_chat_cpp.py
+++ b/examples/low_level_api/low_level_api_chat_cpp.py
@@ -96,7 +96,7 @@ specified) expect poor results""", file=sys.stderr)
 
 		print(file=sys.stderr)
 		print(f"system_info: n_threads = {self.params.n_threads} / {cpu_count()} \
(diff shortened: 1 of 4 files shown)


abetlen:
Thanks @SagsMug, I'll take a look at this and weigh against #55 I think this better preserves streaming functionality.

So is this bug just due to the fact that the Llama vocabulary includes token which are not valid utf-8 strings?

SagsMug:
> Thanks @SagsMug, I'll take a look at this and weigh against #55 I think this better preserves streaming functionality.
> 
> So is this bug just due to the fact that the Llama vocabulary includes token which are not valid utf-8 strings?

According to this:
https://github.com/ggerganov/llama.cpp/pull/73#issuecomment-1465469224
It does indeed stem from some characters needing multiple tokens.

They supposedly fixed this by changing the model converter in https://github.com/ggerganov/llama.cpp/pull/79
But given we still encounter this issue, maybe not.

```
> import llama_cpp
> lparams = llama_cpp.llama_context_default_params()
> ctx = llama_cpp.llama_init_from_file("./models/7B/ggml-model-q4_0.bin", lparams)
> def _tokenize(prompt, bos=True):
 _arr = (llama_cpp.llama_token * (len(prompt) + 1))()
 _n = llama_cpp.llama_tokenize(ctx, prompt.encode("utf8"), _arr, len(_arr), bos)
 return _arr[:_n]
> _tokenize("😀", False)
llama_tokenize: too many tokens
[]
```

I would speculate that the only real solution is to save invalid tokens until a valid output occurs.
But this works alright until that happens.

SagsMug:
> Thanks @SagsMug, I'll take a look at this and weigh against #55 I think this better preserves streaming functionality.
> 
> So is this bug just due to the fact that the Llama vocabulary includes token which are not valid utf-8 strings?

I have added a check that detects a multi-byte UTF8, and waits accordingly.
It

> Thanks @SagsMug, I'll take a look at this and weigh against #55 I think this better preserves streaming functionality.
> 
> So is this bug just due to the fact that the Llama vocabulary includes token which are not valid utf-8 strings?

I have tried to address UTF8 properly by detecting multibytes and waiting for their completion.
Please check that it's satisfactory.
I have also kept the errors = "ignore" as a precaution.

abetlen:
@SagsMug can we reduce the use of `errors="ignore"` to just bare minimum to catch detokenize issues, ie. we probably don't need it for prompts and such.

SagsMug:
> @SagsMug can we reduce the use of `errors="ignore"` to just bare minimum to catch detokenize issues, ie. we probably don't need it for prompts and such.

I have removed a bunch of cases and add a test for this.
Im pretty sure the rest are necessary, but please check

abetlen (approved):

mozzipa:
Because of errors from Llama.generate() and low level api, I use below code snippet. 
How about to integrate it?

```
                if not input_noecho:
                    for id in tokens:
                        detoken = llama_cpp.Llama.detokenize(self=self,tokens=[id])
                        byte_list.append(detoken)
                        gen_tokens += 1
                        try:
                            combine = b''.join(byte_list)
                            letter = combine.decode("utf-8")
                            # print(letter, end="",flush=True)
                            # yield print(letter, end="",flush=True)
                            yield letter
                            if gen_tokens > n_predict:
                                gen_tokens = 0
                                # llama_cpp.llama_free()
                                break
                            byte_list = []
                            
                        except:
                            pass
```

PR #119, opened 2023-04-26
Adding env vars for settings

snxraven:
This allows for most of the settings to be changed using environment variables which would make automatic configuration a lot easier.

This also defaults n_threads to 4 which seems to be the global default I see around repos.

### DIFF
--- a/llama_cpp/server/__main__.py
+++ b/llama_cpp/server/__main__.py
@@ -27,17 +27,16 @@ from sse_starlette.sse import EventSourceResponse
 
 class Settings(BaseSettings):
     model: str
-    n_ctx: int = 2048
-    n_batch: int = 512
-    n_threads: int = max((os.cpu_count() or 2) // 2, 1)
-    f16_kv: bool = True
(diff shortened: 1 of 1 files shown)


snxraven:
The main concern here would be n_threads and max_tokens.

I am hanging out with the https://serge.chat/ group and we noticed that we need to go in and edit max_tokens directly in the file.

With the current code live on the repo is max tokens able to be set as a variable?

I cannot find the references to the 4 Core settings being default I have seen in the past so that may have changed in llama.cpp in the past so we can probably just leave as is if it can be changed via variable

abetlen:
You should be able to `export N_THREADS=4` before running the server which sets it accordingly. As for max_tokens, this is a request parameter so it can be set as in https://platform.openai.com/docs/api-reference/completions/create to whatever you'd like as long as the `prompt + max_tokens < n_ctx`.

snxraven:
Thanks this clears up a ton of questions. We can close this

PR #125, opened 2023-04-29
Make app server module importable

Stonelinks:
First of all thanks for the awesome library!

I don't want to spam you with a bunch of stuff from my fork but I'll try and keep changes minimal and submit PRs, feel free to keep or ignore whichever ones you see fit.

This one just makes the fastapi server `app` importable as a module (see https://github.com/Stonelinks/llama-cpp-python/issues/1 for more details)

PR #126, opened 2023-04-29
Deprecate example server

Stonelinks:
Split this in case these two implementations exist for a reason, but this depends on https://github.com/abetlen/llama-cpp-python/pull/125. Really its just the one commit at the end.

I found it confusing that two fastapi servers exist in this repo (more detail https://github.com/Stonelinks/llama-cpp-python/issues/2)

So this PR "deprecates" the example by pointing folks at the module one, but still remaining runnable.

### DIFF
--- a/examples/high_level_api/fastapi_server.py
+++ b/examples/high_level_api/fastapi_server.py
@@ -4,259 +4,34 @@ To run this example:
 
 ```bash
 pip install fastapi uvicorn sse-starlette
-export MODEL=../models/7B/ggml-model.bin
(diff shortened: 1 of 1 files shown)


jmtatsch:
Found the second server confusing as well. And code duplication.

abetlen:
@Stonelinks I've updated the app server to use a more standard `create_app` pattern, can you update this example then I'm good to merge this in.

Stonelinks:
cool, yeah create_app is better than whatever i did!

fixed up the last commit on this branch

PR #130, opened 2023-04-30
Better server params and fields

Stonelinks:
This is a lot, but the net is that:

- Unsupported fields were removed
- Supported (but missing) fields were added
- A ton of information in parameters were added (via `Field` annotations)

Reasoning should be in commit messages for individual decisions, with some more details here: https://github.com/Stonelinks/llama-cpp-python/issues/3

The result of all this should be better documentation from the swagger / openapi docs available from the server, as well as improved clients generated from `openapi.json` (up next)

image


gjmulder:
Upvote. Muchas gracias! :+1:

abetlen:
@Stonelinks great work!

I want to merge this in but I'd like to keep the unsupported fields in the API as they were before. Otherwise I'm very happy with the addition of the `Fields` object because it works really well with how I've implemented CLI options for the server.

PR #135, opened 2023-05-01
Developer setup

Stonelinks:
I have no idea if my setup is right, the README says "install the package in development mode" but just provides instructions to clone the repo, init submodules, then call setup.py (which will fail because skbuild isn't installed) to compile llama.cpp.

I'm left wondering how to actually set up this project for development. After random stuff breaking / conflicting with my system, I made a stab at making this more explicit with poetry & a virtualenv.

Feedback always welcome


Stonelinks:
hmmmm there might be a problem with this, i assumed `poetry install --all-extras` would pickup setup.py's server extra and install fastapi but that doesn't seem to be the case:

```
(llama-cpp-python-py3.10) humbaba:llama-cpp-python lucas.doyle$ MODEL=./vendor/llama.cpp/models/ggml-vocab.bin VOCAB_ONLY=true python3 -m llama_cpp.server
  File "/Users/lucas.doyle/.pyenv/versions/3.10.11/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/lucas.doyle/.pyenv/versions/3.10.11/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/lucas.doyle/Dropbox/workspace/llama-cpp-python/llama_cpp/server/__main__.py", line 22, in <module>
    from fastapi import Depends, FastAPI
ModuleNotFoundError: No module named 'fastapi'
```

Maybe direct people to use pip instead?

Stonelinks:
i guess poetry extras and setup.py extras are two different things, poetry seems to maintain them in pyproject.toml: https://python-poetry.org/docs/pyproject/#extras

solution: going to add a section for the server there too i guess, meanwhile will give people the option of using pip

Stonelinks:
fixed up the last commit to handle pip and / or poetry

snxraven:
@Stonelinks thanks for this, I have been battling this and installing things manually

abetlen:
@Stonelinks thanks for the contribution, sorry this took so long to merge, appreciate the hard work.

Issue #122, opened 2023-04-27
Builds are slow on MacOS with Apple silicon


zeitderforschung:
Here are the different system infos after loading a model:

**llama.cpp**
```
system_info: n_threads = 4 / 12 | AVX = 0 | AVX2 = 0 | AVX512 = 0 | AVX512_VBMI = 0 | AVX512_VNNI = 0 | FMA = 0 | NEON = 1 | ARM_FMA = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | VSX = 0 |
```

* BLAS, NEON, ARM_FMA and FP16_VA is enabled

**llama-cpp-python**
```
AVX = 0 | AVX2 = 0 | AVX512 = 0 | AVX512_VBMI = 0 | AVX512_VNNI = 0 | FMA = 0 | NEON = 0 | ARM_FMA = 0 | F16C = 0 | FP16_VA = 0 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 1 | VSX = 0 | 
```

* BLAS and SSE3 is enabled, but NEON, ARM_FMA is missing 🤔

matthoffner:
I had a similar issue and upgrading fixed where I see `NEON = 1` and `ARM_FMA = ` now.

`FORCE_CMAKE=1 pip install --upgrade llama-cpp-python`

zeitderforschung:
@matthoffner I unfortunately had no success with your tip with the current version.

zeitderforschung:
I reinstalled with `pip install -U git+https://github.com/abetlen/llama-cpp-python.git`, but without the `FORCE_CMAKE=1` flag and now it works as expected on Apple silicon. Be sure that the command `uname -m` outputs "arm64".

PR #137, opened 2023-05-01
Load dynamically build libs for Apple Platforms (M1/2)

s1lvester:
On Apple Platforms (including ARM64 M1/2) libs are build dynamically, as per CMakeLists.txt. 

Therefore, they get a different extension that needs to be loaded in llama_cpp.py.

Tested on Apple M1 and M2 Platforms (aarch64) with Apple clang version 14.0.0 (clang-1400.0.29.202)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,31 @@
-set(CMAKE_SYSTEM_PROCESSOR "arm64")
-
 cmake_minimum_required(VERSION 3.4...3.22)
 
 project(llama_cpp)
 
-set(BUILD_SHARED_LIBS "On")
-add_subdirectory(vendor/llama.cpp)
-install(
-    TARGETS llama 
-    LIBRARY DESTINATION llama_cpp
-    RUNTIME DESTINATION llama_cpp
-)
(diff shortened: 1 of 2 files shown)


s1lvester:
OK, so here is some more info on why this might possibly be necessary:

M1 / M2 Macs come with Rosetta, an X86 emulation layer, to be able to run legacy macOS binaries. Dynamic linking enables the arm64 libs to be used from inside an x86 emulated terminal.

abetlen:
@s1lvester so the reason I originally defaulted to make + `.so` libraries for MacOS was because many folks reported issues with the `.dylib` based builds with the cmake based build on M1 macs (I think this has to do with the hardware emulation being mismatched between compile and runtime).

s1lvester:
@abetlen oh, I see, I didn't realise this was intentional. I'll close this PR then. 

However I would suggest updating the README to instruct users of Apple Silicon type Platforms to run `arch`  and check that the output is `arm64` before pip-installing llama-cpp-python. Otherwise the llama-cpp will be built for the wrong architecture... 

See also: https://github.com/abetlen/llama-cpp-python/issues/122#issuecomment-1528770790

PR #146, opened 2023-05-03
added huggingface space implementation

abhishekmamdapure:


winglian:
@abhishekmamdapure  seems broken, does it actually work?

Issue #141, opened 2023-05-02
Update low-level examples to use new sampling API


abetlen:
llama.cpp recently changed the sampling API to make sampling more configurable. Unfortunately this broke many of the low-level examples so we should update these.

mikeggh:
I'd like to know an example for using:
llama_load_session_file(ctx, session_file, embd, out_len, (llama_cpp.c_int * out_len)(*embd_inp_out))

and save of course... if you get a chance :> i'll try to brute force it later though..

SagsMug:
Im gonna work on updating the chat example, it might take a moment given the number of changes 😉

abetlen:
Thanks @SagsMug yeah I was a little overwhelmed with the new sampling API but overall it seems like the right direction to have more control over sampling. Check out https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama.py#L255 might help

PR #153, opened 2023-05-04
Update low_level_api examples

SagsMug:
Updates chat example
Updates plain llama example
Adds some usage examples

TODO:
- Test lora

Should fix https://github.com/abetlen/llama-cpp-python/issues/141


SagsMug:
@abetlen I think you can review this now.
I dont really know any lora's, so its left untested
Mirostat was only added to chat, i can add it to the plain example aswell.

abetlen:
@SagsMug looks good to me, great work.

PR #157, opened 2023-05-05
readme windows

th-neu:
added the simple run server command for the windows cmd to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,9 +64,17 @@ This allows you to use llama.cpp compatible models with any OpenAI compatible cl
 
 To install the server package and get started:
 
+Linux
 ```bash
 pip install llama-cpp-python[server]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,12 +64,20 @@ This allows you to use llama.cpp compatible models with any OpenAI compatible cl
 
 To install the server package and get started:
 
+Linux
```
abetlen, line 67: I believe this also covers MacOS, so should be `Linux / MacOS`

abetlen (changes requested):

README.md
th-neu: yeah wasn't sure. No Mac here

abetlen (approved):

PR #172, opened 2023-05-08
Add truncate to high level api

SagsMug:
This PR implements 3 optional truncation strategies.
- Start, which removes tokens from the start of the prompt, except the BOS token. 
  Useful for story generation etc. 
- Middle, which removes from the middle. 
  Useful for chatbots.
- End, which removes from the end. 
  I dont really know why this is used, but its commonly implemented.


SagsMug:
Hey @abetlen 
I know you might be busy, but if you would consider checking out this commit, i would be grateful!

PR #176, opened 2023-05-09
WIP: Mechanism to retrieve all logprobs on completion

tristanvdb:
Hi All,

I have been working on a simple "sampling" algorithm for LLM. It is very simple:
> given a prompt and a list of string, which string is the most likely completion?

I included a notebook with a greedy implementation of a `Choice` algorithm. I demonstrated a couple use-cases using a "raw" LLama 7B model (4bits quant.).

My changes to `LLama._create_completion` look a bit hacky. Instead, it might be better to have: `LLama.greedy(prompt)` which returns all logprobs for the next token. It will let us play with search algorithms (good old AI).

I am submitting that here to get feedbacks on:
 - API for greedy completion: does my change work? or should we have `LLama.greedy(prompt)`?
 - Choice algorithm:
    - I use it in a similar way as the ReAct-like example in the notebook: Coming soon...
    - It should eventually be implemented in LLama.cpp instead but `python == fast prototyping`

Thanks,
-Tristan


abetlen:
@tristanvdb thanks for the PR, I think one of the next pieces of development is going to be to make the high-level API a little more functional / modular to make implementing features like this a little easier.

tristanvdb:
I just released [AutoCog](https://github.com/LLNL/AutoCog) which uses this choice algorithm to implement a programming language that is compiled to be executed by the LM. AutoCog uses my fork of llama-cpp-python to access LLama.cpp. I'd prefer it to use the mainline.

@abetlen Any progress on this kind of features?

PS: Paper link through [Hacker News](https://news.ycombinator.com/item?id=36413084)

PR #177, opened 2023-05-09
Updated installation instructions for BLAS backends

joelkurian:
llama.cpp does not read BLAS related environment variables while configuring build with CMake. 

Instead, it needs to be passed as command line argument which correctly configures BLAS backend. It can be done by setting `CMAKE_ARGS` environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,19 +35,19 @@ Use the `FORCE_CMAKE=1` environment variable to force the use of `cmake` and ins
 To install with OpenBLAS, set the `LLAMA_OPENBLAS=1` environment variable before installing:
 
 ```bash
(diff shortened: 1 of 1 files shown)


abetlen:
@joelkurian what's your OS? Happy to add these instructions but the environment variables do work for me but I'm on Linux.

joelkurian:
I am on ArchLinux. Environment variables work with `make` build, but have a bit weird behavior with `cmake`.

In order to make use of environment variable, CMake needs to read it like `$ENV{LLAMA_CUBLAS}` in CMakeLists.txt in llama.cpp which is not the case right now.
https://github.com/ggerganov/llama.cpp/blob/master/CMakeLists.txt

Issue #171, opened 2023-05-08
Base settings confusion (How to disable mlock)


snxraven:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [X] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [X] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [X] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [X] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior
Before the latest updates I have been using ENV Vars to set base settings for things like cache, n_threads etc.

This does not seem to work anymore, I noticed the readme was updated to show the use of switches like this:
--model models/7B/ggml-model.bin

Is this how all base settings are given now? The model switch above does infact work.

Usually llama.cpp will default to mmap by itself but in the new versions I keep seeing the cannot mlock errors.

I have attempted ENV and switches to disable mlock and I cannot seem to do so.

I have attempted:

--use_mlock 0

--use_mmap 1

a combination including both of these togetjer and seperate.

While testing was happening the env vars were not touched.

# Current Behavior

Instead of disabling mlock, it seems that its default every time I run the server.

abetlen:
This is a bug, the behaviour is that it should allow for both environmnet variables and cli options and validate at the end.

@gjmulder for dockerfiles the environmnet variables are still preferred because they play better with the variaous container deployment methods

abetlen:
@snxraven a fix should be pushed that lets you use the environment variables, let me know if that works. You can also call `--help` now for a full list of cli arguments and their defualt values.

snxraven:
@abetlen I have attempted the latest it does not seem to fall back for me and I still see the memlock errors:
![image](https://user-images.githubusercontent.com/1396275/236904240-3119adac-0f18-4e60-96bc-82c01dafd42f.png)

All of my previous settings within my docker env are ignored in 0.1.46

I confirmed this because cache and n_threads was also ignored.

gjmulder:
`MODEL` does work, however.

Ditto on the `MLOCK` warning. I worked around it with the following change to my `Dockerfile`:

[...]

PR #190, opened 2023-05-12
Update README.md to reflect changes for passing the model for the docker image

th-neu:
Since its now --model the old way of -emodel does not work anymore.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "llama_cpp_python"
-version = "0.1.49"
+version = "0.1.50"
 description = "Python bindings for the llama.cpp library"
 authors = ["Andrei Betlen <abetlen@gmail.com>"]
 license = "MIT"
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ setup(
     description="A Python wrapper for llama.cpp",
     long_description=long_description,
     long_description_content_type="text/markdown",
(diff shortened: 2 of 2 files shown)


abetlen:
@th-neu the environment variables should still work as a fallback to cli options, there was a bug with this but should've been fixed in #171

th-neu:
The entrypoint is not an array. As such passing anything will not work afaik. I am on that

abetlen:
Ah gotcha, thanks!

Just to confirm though, does the `-e MODEL` approach work for you?

th-neu:
> Ah gotcha, thanks!
> 
> Just to confirm though, does the `-e MODEL` approach work for you?

No. Always getting "none is not an allowed value (type=type_error.none.not_allowed)" as in the model is not found.
Docker version 23.0.6, build ef23cbc

abetlen:
I think that's an older version can you try running `docker images` then `docker rmi` by the image you have by id and pull it again.

th-neu:
> I think that's an older version can you try running `docker images` then `docker rmi` by the image you have by id and pull it again.
> 
> I'll try to update the docker publish action to use the release tags.

I am building my own docker images. And trying yours. So those are up to date. Does it work for you? A typo for me would not be uncommen. Building the images is not the problem. Running is.

th-neu:
#191 this should fix the passing of the Model. This README.md update would then not matter anymore

PR #191, opened 2023-05-12
chore: Update Dockerfile

th-neu:
Setting workdir so image is a bit cleaner. Setting a model variable. Passing a model env variable to the run command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,9 @@ FROM python:3-slim-bullseye
 # We need to set the host to 0.0.0.0 to allow outside access
 ENV HOST 0.0.0.0
 
-COPY . .
+WORKDIR /app
+
+COPY . /app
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,13 +3,18 @@ FROM python:3-slim-bullseye
...
 RUN python -m pip install --upgrade pip pytest cmake scikit-build setuptools fastapi uvicorn sse-starlette
 
 RUN LLAMA_OPENBLAS=1 python3 setup.py develop
 
+# Setting a env variable MODEL
+ENV MODEL ggml.bin
+
 # Run the server
-CMD python3 -m llama_cpp.server
+CMD python3 -m llama_cpp.server --model $[MODEL}
```
jmtatsch, line 20: $[MODEL} looks wonky

PR #198, opened 2023-05-12
Allow relative paths at model initialization

andreakiro:
Loading models requires absolute path right now. This change allows for the use of relative (and environment variables)  paths, increasing flexibility and ease of use. This could be especially beneficial in downstream libraries usage, such as `langchain` and many others, where users may load models from various directories.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -124,7 +124,7 @@ class Llama:
             A Llama instance.
         """
         self.verbose = verbose
-        self.model_path = model_path
+        self.model_path = os.path.abspath(os.path.expanduser(os.path.expandvars(model_path)))
 
         self.params = llama_cpp.llama_context_default_params()
         self.params.n_ctx = n_ctx
@@ -149,8 +149,8 @@ class Llama:
         self.lora_base = lora_base
         self.lora_path = lora_path
 
-        if not os.path.exists(model_path):
-            raise ValueError(f"Model path does not exist: {model_path}")
+        if not os.path.exists(self.model_path):
+            raise ValueError(f"Model path does not exist: {self.model_path}")
 
         self.ctx = llama_cpp.llama_init_from_file(
             self.model_path.encode("utf-8"), self.params


andreakiro:
@abetlen hi, do you have a feedback on this one? :)

gjmulder:
#209?

abetlen:
@andreakiro what's your operating system? I only ask because relative paths have always worked for me.

andreakiro:
@abetlen Here is a small script that illustrate the problem on my machine:

```python
from llama_cpp import Llama
print(__file__) # "~/desktop/opensource/llama-binding/example.py"

# -- with relative paths

# binary is child of project root (relative path)
model_path = "./models/ggml-model-q4_0.bin"
Llama(model_path=model_path) # this works fine 

# binary is somewhere else on disk (relative path)
model_path = "~/.cache/models/ggml-model-q4_0.bin"
Llama(model_path=model_path) # this fails
# ValueError: Model path does not exist: ~/.cache/models/ggml-model-q4_0.bin

# -- with absolute paths

# binary is somewhere else on disk (absolute path to bin)
model_path = "/Users/username/.cache/models/ggml-model-q4_0.bin"
Llama(model_path=model_path) # this works fine again
```

I am running this on an Apple M1 chip (Darwin Kernel 22.4.0, arm64), with this Python env:

```
python==3.11.2
llama-cpp-python==0.1.49
```

abetlen:
@andreakiro ah gotcha. So strictly speaking that isn't a relative path, it's a shortcut that's expanded through an environment environment variable, this is usually done by the user shell (for example when running this from my cli it expands correctly).

My issue with merging this is in is that it's non-standard to do this expansion automatically inside a library, it means the path the user passes in is going to be incompatible with most other libraries (for example most of the the path arguments in the python standard lib) that expect fully expanded paths which feels like a footgun.

andreakiro:
@abetlen Hm I see. I agree that in general it should be user responsibility to provide a valid path. But I am not sure to understand why this non-standard in-library automatic expansion could lead to further incompatibility here..? Given a path (valid or including non-expanded shortcuts) `llama-cpp-python` only use it to load the model binary right?

PR #240, opened 2023-05-18
Add `min_tokens` argument

Yard1:
This PR adds a `min_tokens` argument to complement `max_tokens`. When set, the EOS token will be discarded until `min_tokens` have been generated.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -598,6 +598,7 @@ class Llama:
         prompt: str,
         suffix: Optional[str] = None,
         max_tokens: int = 16,
+        min_tokens: int = 0,
         temperature: float = 0.8,
         top_p: float = 0.95,
         logprobs: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


Yard1:
@abetlen, could you review when you have a moment? Thank you!

abetlen:
Hey @Yard1 sorry to take so long on this reply. The issue I see with this at the moment is that the stop token is still being generated and appended to the eval_logits and eval_tokens internally. This will probably cause some kind of issue in generation, what you really want is a min_tokens inside of generate or something that essentially ignores / sets the eos token probability to 0 until a certain number of tokens are generated.

Also, it should probably be noted that fewer than min_tokens may be returned if e.g. another stop criteria like a stop sequence is encountered.

Yard1:
Got it, thanks! Let me take a look at this.

Yard1:
Hmm, now that I think about it, this can be easily implemented through a `LogitsProcessor`. Therefore, we don't need a separate argument (and by extension, this PR). Let me know if that makes sense - we can close this PR then.

IkariDevGIT:
any updates on this?

abetlen:
@Yard1 we've added the parameter to the server and it's available in the python API via a new `MinTokensLogitProcessor`

PR #246, opened 2023-05-19
feat: make embedding support list of string as input

chabotsi:
The /v1/embeddings route of OpenAI api allows a string or a list of string as input.
This updates allows the of /v1/embeddings to mimic the OpenAI's one.


abetlen:
Thanks @chabotsi

PR #249, opened 2023-05-20
Dockerfile optimization

localagi:
use entrypoint to allow running the container like a command

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,4 +12,5 @@ RUN python -m pip install --upgrade pip pytest cmake scikit-build setuptools fas
 RUN LLAMA_OPENBLAS=1 python3 setup.py develop
 
 # Run the server
-CMD python3 -m llama_cpp.server
+ENTRYPOINT ["python3","-m","llama_cpp.server"]
+CMD ["--model","models/7B/ggml-model.bin"]
--- a/Dockerfile.cuda
+++ b/Dockerfile.cuda
@@ -13,4 +13,5 @@ RUN python3 -m pip install --upgrade pip pytest cmake scikit-build setuptools fa
 RUN LLAMA_CUBLAS=1 python3 setup.py develop
 
 # Run the server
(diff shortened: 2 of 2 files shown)


gjmulder (changes requested):
The [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md) says to pass the model file this way:

`-e MODEL=/models/ggml-model-name.bin`

@abetlen mentioned somewhere there is a priority over whether the env `$MODEL` or the arg `--model` overrides the other, but I can't find it anywhere in the docs.

In any case, having the model specified differently with two different paths in two different places is confusing to newcomers. The `README.md` needs to be updated to reflect these changes to the Dockerfiles.

I personally prefer the current `$MODEL` approach as it means one can easily switch models via `docker run` without any default model path that might throw confusing errors.

localagi:
> The [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md) says to pass the model file this way:
> 
> `-e MODEL=/models/ggml-model-name.bin`
> 
> @abetlen mentioned somewhere there is a priority over whether the env `$MODEL` or the arg `--model` overrides the other, but I can't find it anywhere in the docs.
> 
> In any case, having the model specified differently with two different paths in two different places is confusing to newcomers. The `README.md` needs to be updated to reflect these changes to the Dockerfiles.
> 
> I personally prefer the current `$MODEL` approach as it means one can easily switch models via `docker run` without any default model path that might throw confusing errors.

Did you try the approach you mentioned? Setting the model as env variable - I don't know if that would even work with the latest dockerfile. Maybe MODEL is some environment var of llama-cpp-python itself....

Entypoint and CMD have a delicate way of working together.

abetlen:
@localagi I believe @gjmulder is correct in that specifying a default may lead to some issues. `llama-cpp-python` uses pydantic `BaseSettings` to parse configuration variables as well as CLI options.

PR #252, opened 2023-05-20
Added and implemented logits processor function list and stop criteria function list.

Maximilian-Winter:
I added and implemented logits processor function list and stop criteria function list for use with microsoft guidance.


mmagnesium:
This is going to be really helpful. Do you need any help with this?

Maximilian-Winter:
@mmagnesium I have implemented them basically

PR #258, opened 2023-05-22
Change docker build dynamic param to image instead of cuda version

Pipboyguy:
Give user option to specify exact CUDA image for more dynamic builds


gjmulder (changes requested):
Good change as it for a bit more flexibility, however you've pulled in `vendor/llama.cpp`

Pipboyguy:
@gjmulder Thanks for the spot! Fixed it hopefully all looks good now.

gjmulder (approved):
LGTM

PR #271, opened 2023-05-24
Implemented logit processors and stop criteria's

Maximilian-Winter:
I made a implementation of logit processors and custom stop criteria processors.


Maximilian-Winter:
They are needed for proper guidance support:
https://github.com/microsoft/guidance/pull/70

llama_cpp/llama.py
```diff
@@ -316,9 +316,14 @@ def _sample(
         mirostat_tau: llama_cpp.c_float,
         mirostat_eta: llama_cpp.c_float,
         penalize_nl: bool = True,
+        logits_processors=None
```
abetlen, line 319: Add type signature

llama_cpp/llama.py
```diff
@@ -328,6 +333,10 @@ def _sample(
             else last_n_tokens_size
         )
         logits = self.eval_logits[-1]
+        for processor in logits_processors:
+            logits = processor(list(self.eval_tokens), logits)
+
+        self.eval_logits[-1] = logits
```
abetlen, line 339: Does logits_processor have to modify the logits, can it simply update the logits array here during the sampling?

llama_cpp/llama.py
```diff
@@ -436,6 +445,8 @@ def sample(
         mirostat_eta: float = 0.1,
         mirostat_tau: float = 5.0,
         penalize_nl: bool = True,
+        logits_processors=None
```
abetlen, line 448: Add type signature

llama_cpp/llama.py
```diff
@@ -484,6 +497,7 @@ def generate(
         mirostat_mode: int = 0,
         mirostat_tau: float = 5.0,
         mirostat_eta: float = 0.1,
+        logits_processors=None
```
abetlen, line 500: Add type signature

abetlen, line 500: Can we move `stopping_criteria` here so it matches the hugginface `generate` API? Then only use `generate` in the guidance PR.

abetlen (changes requested):

llama_cpp/llama.py
Maximilian-Winter: Yes the logits does have to modify the logits.

llama_cpp/llama.py
Maximilian-Winter: I added a type signature.

llama_cpp/llama.py
Maximilian-Winter: I added a type signature.

Maximilian-Winter: I would have to move the stop criteria check down here. Right now I'm using create completion in guidance.

llama_cpp/llama.py
```diff
@@ -445,7 +444,7 @@ def sample(
         mirostat_eta: float = 0.1,
         mirostat_tau: float = 5.0,
         penalize_nl: bool = True,
-        logits_processors=None
+        logits_processors:  List[Callable[[List[llama_cpp.c_int], List[llama_cpp.c_float]], List[float]]] = None
```
abetlen: Name should be changed to `logits_processor` to match hugginface generate argument in https://huggingface.co/docs/transformers/v4.26.0/en/main_classes/text_generation#transformers.GenerationMixin.generate.logits_processor

Additionally, the types here should be changed to python `float` and `int`

abetlen (changes requested):

PR #277, opened 2023-05-26
Use numpy for internal buffers

abetlen:
Replace internal buffers with numpy arrays, should reduce memory usage and inference time as numpy buffers can be passed directly to c.


abetlen:
One potential gotch we should watch for is that numpy structured arrays are not necessarily aligned like the corresponding c-structures https://numpy.org/doc/stable/user/basics.rec.html#automatic-byte-offsets-and-alignment

Not sure what the general solution to this would be however.

Issue #312, opened 2023-06-04
Mirostat mode -- Possibly missing parameters or unclear usage


CoffeeVampir3:
I'm not sure if this I'm missing how to initialize mirostat mode, however it seems like it's not currently possible to use:

```python
    def __init__(
        self,
        model_path: str,
        # NOTE: These parameters are likely to change in the future.
        n_ctx: int = 512,
        n_parts: int = -1,
        n_gpu_layers: int = 0,
        seed: int = 1337,
        f16_kv: bool = True,
        logits_all: bool = False,
        vocab_only: bool = False,
        use_mmap: bool = True,
        use_mlock: bool = False,
        embedding: bool = False,
        n_threads: Optional[int] = None,
        n_batch: int = 512,
        last_n_tokens_size: int = 64,
        lora_base: Optional[str] = None,
        lora_path: Optional[str] = None,
        verbose: bool = True,
    ):
 ```
 
 In the init, there's no mention of mirostat_mode, however in the sample and generate methods, it does exist:
 ```python
    def _sample(
        self,
        last_n_tokens_data,  # type: llama_cpp.Array[llama_cpp.llama_token]
        last_n_tokens_size: llama_cpp.c_int,
        top_k: llama_cpp.c_int,
        top_p: llama_cpp.c_float,
        temp: llama_cpp.c_float,
        tfs_z: llama_cpp.c_float,
        repeat_penalty: llama_cpp.c_float,
        frequency_penalty: llama_cpp.c_float,
        presence_penalty: llama_cpp.c_float,
        mirostat_mode: llama_cpp.c_int,
        mirostat_tau: llama_cpp.c_float,
        mirostat_eta: llama_cpp.c_float,
        penalize_nl: bool = True,
        logits_processor: Optional[LogitsProcessorList] = None,
    ):
 ```
 
 I didin't see anything in the code that would toggle mirostat on after the fact. In LlamaCpp these are enabled by passing:
 
 ```
  --mirostat N use Mirostat sampling.

Top K, Nucleus, Tail Free and Locally Typical samplers are ignored if used.

(default: 0, 0 = disabled, 1 = Mirostat, 2 = Mirostat 2.0)

--mirostat_lr N Mirostat learning rate, parameter eta (default: 0.1)

--mirostat_ent N Mirostat target entropy, parameter tau (default: 5.0). 
```

It seems like the initialization function is maybe just missing these definitions, but I wasn't sure if I was missing a correct method of activating the mode.

[...]

PR #329, opened 2023-06-06
Added Mirostat Mode and related Params to Llama initialization

CoffeeVampir3:
This change is mostly motivated by these parameters being similar to top-k and temperature, which are present in the Llama initialization. This should make utilizing these parameters more user friendly and more consistent with LlamaCpp's internal api.

See issue #312 for some additional context.


CoffeeVampir3:
Added a fix to langchain to make this work as well: <https://github.com/CoffeeVampir3/langchain>

Holding off on the langchain PR until this one is solidified :+1:

abetlen:
Hey @CoffeeVampir3 sorry to take so long to look at this. As I see it your PR adds the mirostat parameters as instance parameters, however mirostat sampling is currently possible through both the server and __call__ API, is there a reason this change to be implemented?  Are you looking to override the sampling of an existing client you don't have control over?

CoffeeVampir3:
> Hey @CoffeeVampir3 sorry to take so long to look at this. As I see it your PR adds the mirostat parameters as instance parameters, however mirostat sampling is currently possible through both the server and **call** API, is there a reason this change to be implemented? Are you looking to override the sampling of an existing client you don't have control over?

Not at all, thanks very much for your hard work on the library.

The reason for this was motivated by my work with langchain, which adapts over llama-cpp-python. If I were using llama-cpp, I'd pass in the command line parameters --mirostat_mode 2, --mirostat_tau .9, etc. With this set up in the initializer, you get quite a clean api that is consistent with llama-cpp itself:

```
llm = LlamaCpp(
    model_path=os.path.abspath(sys.argv[1]),
    lora_path=os.path.abspath(sys.argv[2]) if len(sys.argv) > 2 else None,
    n_batch=512,
    n_ctx = 4_096,
    n_gpu_layers = 60,
    max_tokens = 2048,
    temperature = 1.5,
    top_k = 0,
    top_p = .95,
    verbose = False,
    stop = ("Human:", "AI:"),
    mirostat_mode = 2,
    mirostat_tau = 5.0,
    mirostat_eta = 0.1,
)
```

This is both an aethsetic choice but also a practical one, as the usage of llama-cpp-python as an adapter becomes quite a bit more difficult to enable this parameter without this change. For example, one method would be to directly override langchains run or generate and inject the parameters into the run call. This gets tricky, partially because of the design of langchain, but also it creates an inconsistency where the mirostat parameters all require special handling and special code. An example hack I wrote to inject the parameters that lets you handle things somewhat consistently with other parameters is this, but this feels very hacky and not really needed:

```
    def _get_parameters(self, stop: Optional[List[str]] = None) -> Dict[str, Any]:
        #...
        params = self._default_params

        # llama_cpp expects the "stop" key not this, so we remove it:
        params.pop("stop_sequences")

        # then sets it as configured, or default to an empty list:
        params["stop"] = self.stop or stop or []

        params["mirostat_mode"] = 2
        params["mirostat_eta"] = 0.1
        params["mirostat_tau"] = 5.0
        return params
```

To summarize, this doesn't give any new capabilities but I think makes the overall api easier to use and more consistent in how the parameters are handled.

nahuel89p:
Is it just me or this implementation doesn't change the outputs at all? No matter how I tweak tau and eta, the outputs are unchanged.

PR #370, opened 2023-06-13
fix: Make LLamaState pickleable for disk cache

Okabintaro:
I fixed the issue by making the saved state a bytes object instead of the ctypes one which can't be pickled.
Not sure if this is ideal though, maybe the copy on loading from cache can be avoided.

Stumbled upon another bug in the disk cache which caused it to crash on pushing keys using the disk cache library.
When using `cache.push()`, an integer gets put into the cache as a key(see documentation) which then won't be comparable to the token sequence on subsequent cache fetches.
I commented that line out for now but there should be a better way to handle this.


gjmulder:
Bump

abetlen (approved):

PR #392, opened 2023-06-17
Use HTTP vs SSH in git clone example

drgnfr6:
What it says on the tin, just some doc cleanup.


SubhranshuSharma:
The readme is not wrong, Git repositories can b cloned over many protocols, it used ssh syntax which ur asking to change to http which is fair enough, just maybe change the title to something like 'using http over ssh in git clone example'

PR #405, opened 2023-06-20
Update README.md

mikeyang01:
"git clone git@github.com:abetlen/..." requires access rights to the git account, causing clone failure. 
```
git clone --recurse-submodules git@github.com:abetlen/llama-cpp-python.git
```


drgnfr6:
See pull #392

PR #415, opened 2023-06-22
server: pass seed param from command line to llama

lexin4ever:


abetlen:
@lexin4ever good catch, hadn't realized this wasn't already specified

Issue #349, opened 2023-06-08
Logprobs incorrectly computed from logits


progtologist:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [X] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [X] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [X] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [X] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

I am feeding llama-cpp-python through the `create_completion` api with predefined prompts.
The predefined prompts are like this:
```
Q: Play music
A: music.play()

Q: Stop music
A: music.stop()

... a few more examples...

Q: Pause music
A: %STRING%
```
The `%STRING%` I replace with different API calls, one of them being `music.pause()` but the others can be `music.play()`, `music.stop()` etc.

I am calling the API with max_tokens=0, logprobs=1, temperature=0 and echo=True. I am basically not interested in generating a response, I just want my `%STRING%` to be evaluated by the LLM.

The reason why I do this, is that I expect the correct API call to have a much higher combined logprob than the wrong API calls.

I have validated this example with OpenAI with different models, and it does work without issues.

The problem that I am facing is that the logprobs that gets calculated from llama-cpp-python does not appear to be correct. 

The values that I get are almost constantly extremely low. I have tried to see if I can fix it, but unfortunately the few things that I have tried did not work, and I believe some expert advice could go a long way.

# Current Behavior

I print all the tokens, together with their logprobs and their probs which I calculate through numpy.exp(logprobs)

The following is an example of the results I get:
```
Token Logprob Probability
mu -13.098522331165873 2.0482550320620133e-06
sic -23.94800838365418 3.976601815073668e-11
. -25.437061074011822 8.970667086442726e-12
play -24.84781513922617 1.6170781080207215e-11
() -27.32543331183062 1.3574226671545055e-12
```

Similarly, the same example with the OpenAI api yields:
```
Token Logprob Probability
mu -0.49815017 0.6076536766979097
sic -0.0052926685 0.9947213129925816
. 0 1.0
play -0.8366768 0.4331475703151241
[...]

PR #453, opened 2023-07-07
Fix incorrect token_logprobs (due to indexing after sorting)

wu-qing-157:
Thanks for the wonderful Python wrapper for llama.cpp.
When using your framework, I found a problem in `output['logprobs']['token_logprobs]`.

For example, on commit ca11673 (latest main branch when I write this) the following code
```python
from llama_cpp import Llama
llama = Llama('.../llama.cpp/models/30B/ggml-model-q8_0.bin', logits_all=True)
print(llama('I', temperature=0, max_tokens=5, logprobs=1))
```
will output `{'id': 'cmpl-caa5cef2-ae2f-4157-8323-2dc644bb6308', 'object': 'text_completion', 'created': 1688724957, 'model': '/data/yi/llama.cpp/models/30B/ggml-model-q8_0.bin', 'choices': [{'text': "'m a big fan", 'index': 0, 'logprobs': {'tokens': ["'", 'm', ' a', ' big', ' fan'], 'text_offset': [1, 2, 3, 5, 9], 'token_logprobs': [-19.985351200841283, -19.405950866756548, -7.967424092851152, -12.165255948926252, -17.411318060240337], 'top_logprobs': [{"'": -2.2789093215688228}, {'m': -0.5416520462609419}, {' a': -2.119851766190996}, {' big': -2.872671052838605}, {' fan': -0.29880118020493784}]}, 'finish_reason': 'length'}], 'usage': {'prompt_tokens': 2, 'completion_tokens': 5, 'total_tokens': 7}}`. You can see the `token_logprobs` are quite small, and are different from the numbers in `top_logprobs`, which are assumed to be consistent.

When I look into the code, I find that this unexpected behavior is caused by indexing the sorted logprobs. In L1134 of https://github.com/abetlen/llama-cpp-python/blob/ca11673061ecd9198b4800f68073ae14d4440ecd/llama_cpp/llama.py#L1123-L1140 since `sorted_logprobs` is already sorted, indexing the list by token id does not produce the corresponding logprobs. We should instead index the list before sorting, like the one in L1139.
Same fix for L961 and L1036.

After the fix, on commit 9e61661, the above code outputs `{'id': 'cmpl-74ba65a7-1978-4dcd-aa43-09b35ec8361c', 'object': 'text_completion', 'created': 1688725240, 'model': '/data/yi/llama.cpp/models/30B/ggml-model-q8_0.bin', 'choices': [{'text': "'m a big fan", 'index': 0, 'logprobs': {'tokens': ["'", 'm', ' a', ' big', ' fan'], 'text_offset': [1, 2, 3, 5, 9], 'token_logprobs': [-2.2789093215688228, -0.5416520462609419, -2.119851766190996, -2.872671052838605, -0.29880118020493784], 'top_logprobs': [{"'": -2.2789093215688228}, {'m': -0.5416520462609419}, {' a': -2.119851766190996}, {' big': -2.872671052838605}, {' fan': -0.29880118020493784}]}, 'finish_reason': 'length'}], 'usage': {'prompt_tokens': 2, 'completion_tokens': 5, 'total_tokens': 7}}`, where the `token_logprobs` matches `top_logprobs`.

I think #349 is also due to this bug.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -958,7 +958,7 @@ class Llama:
                                 )
                             ],
                             "text_offset": [text_offset],
-                            "token_logprobs": [sorted_logprobs[int(token)][0]],
+                            "token_logprobs": [current_logprobs[int(token)]],
                             "top_logprobs": [top_logprob],
                         }
                     returned_tokens += 1
@@ -1033,7 +1033,7 @@ class Llama:
                             self.detokenize([token]).decode("utf-8", errors="ignore")
                         ],
                         "text_offset": [text_offset],
-                        "token_logprobs": [sorted_logprobs[int(token)][0]],
+                        "token_logprobs": [current_logprobs[int(token)]],
                         "top_logprobs": [top_logprob],
                     }
 
@@ -1131,7 +1131,7 @@ class Llama:
                         zip(logprobs_token, range(len(logprobs_token))), reverse=True
                     )
                 )
-                token_logprobs.append(sorted_logprobs[int(token)][0])
+                token_logprobs.append(logprobs_token[int(token)])
                 top_logprob: Optional[Dict[str, float]] = {
                     self.detokenize([i]).decode("utf-8", errors="ignore"): logprob
                     for logprob, i in sorted_logprobs[:logprobs]


matthoffner:
Thanks @wu-qing-157 I'm adding a PR to lm-eval for ggml and hoping this is related to the issues I'm seeing.

abetlen:
@wu-qing-157 thanks for the catch! LGTM

abetlen (approved):

PR #460, opened 2023-07-09
Add support for llama.cpp's --tensor-split parameter

shouyiwang:
The current llama-cpp-python does not include support for --tensor-split parameter. When running a large model across two GPUs, it currently loads the model by default in a half-by-half manner. However, this approach presents certain issues. For example, when a user has two GPUs with different VRAM sizes, it can lead to OOM. Implementing the --tensor-split parameter will address this problem by empowering users to define the proportion of the model distributed across multiple GPUs.

I'm uncertain if importing ctypes into llama.py is the most appropriate approach. However, I'm currently unsure of an alternative solution in llama_cpp.py. I would greatly appreciate any advice or suggestions regarding this matter.

Tested thouroughly with text-generation-webui. I'll sumbit a PR there after this PR get merged. Thx!

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -19,13 +19,13 @@ from typing import (
 from collections import deque, OrderedDict
 
 import diskcache
+import ctypes
 
 from . import llama_cpp
 from .llama_types import *
 
 import numpy as np
 import numpy.typing as npt
-
 
 class BaseLlamaCache(ABC):
     """Base cache class for a llama.cpp model."""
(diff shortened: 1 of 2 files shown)


shouyiwang:
Hi @abetlen ,
I just wanted to kindly draw your attention to this PR that I submitted 5 days ago. It would be great if you could review it when you have some time. I am available to make any necessary changes or answer any questions you might have. 

Thank you for your time and consideration.

abetlen:
@shouyiwang thank you for the contribution, lgtm

shouyiwang:
@abetlen Thank you so much!!

PR #466, opened 2023-07-10
Add bindings for custom_rope

randoentity:
This adds bindings for https://github.com/ggerganov/llama.cpp/pull/2054
Tested with text-generation-webui (will link PR here soon)

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -205,6 +205,8 @@ class Llama:
         model_path: str,
         # NOTE: These parameters are likely to change in the future.
         n_ctx: int = 512,
+        rope_freq_base: float = 10000.0,
+        rope_freq_scale: float = 1.0,
         n_parts: int = -1,
         n_gpu_layers: int = 0,
         seed: int = 1337,
@@ -227,6 +229,8 @@ class Llama:
         Args:
             model_path: Path to the model.
             n_ctx: Maximum context size.
+            rope_freq_base: RoPE base frequency.
+            rope_freq_scale: RoPE frequency scale.
             n_parts: Number of parts to split the model into. If -1, the number of parts is automatically determined.
             seed: Random seed. -1 for random.
             f16_kv: Use half-precision for key/value cache.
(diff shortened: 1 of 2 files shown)


gjmulder:
I am eagerly awaiting this feature, but I think we should wait for the upstream `llama.ccp` to ensure it is stable before incorporating it downstream. As per [my comment here](https://github.com/ggerganov/llama.cpp/discussions/1965#discussioncomment-6395555) this is potentially a major breaking change if not implemented carefully.

randoentity:
I agree that we should wait merging this. Shall I convert it to draft? Even though it wasn't hard to get this working, I like to have this available so more people test/play around with this.
### results
Tested with [vicuna 7b 1.3](https://huggingface.co/TheBloke/vicuna-7B-v1.3-GGML/blob/main/vicuna-7b-v1.3.ggmlv3.q8_0.bin).
```
n_ctx      = 4096
freq_base  = 20000.0
freq_scale = 0.83
```
I generated a key-value array with unique keys and values amounting to roughly 3K tokens and appended `In the above dict what is the value of the key "<first key>"?` to the instruct prompt and answer was correct. I did have to play with the prompt a bit to get consistent results, and mentioning "dict" seemed to do the trick. I used the "Divine Intellect" parameters. Not a substitute for a perplexity test, but I'm quite resource limited.
FWIW I did not enable superbooga.
The result was:
`Output generated in 7.47 seconds (2.95 tokens/s, 22 tokens, context 3160, seed 442613330)`
If I made some wrong assumptions, please let me know; I'm here to learn and hopefully contribute something useful in the future.

Gyztor:
> I am eagerly awaiting this feature, but I think we should wait for the upstream `llama.ccp` to ensure it is stable before incorporating it downstream. As per [my comment here](https://github.com/ggerganov/llama.cpp/discussions/1965#discussioncomment-6395555) this is potentially a major breaking change if not implemented carefully.

Looks like they merged it on llama.cpp as of 6 hours ago now: https://github.com/ggerganov/llama.cpp/pull/2054

PR #486, opened 2023-07-18
Fix #484, add DRY_RUN to not build any library

### LINKED ISSUE #484: Can it provide an cmake option to disable building of libllama.so?
Can it provide a cmake option to disable building of libllama.so? I hope use `/usr/lib/libllama.so` provided by [llama.cpp](https://github.com/ggerganov/llama.cpp) not `/usr/lib/python3.11/site-packages/llama_cpp/libllama.so`. I am packaging it to linux distribution, if so, user will not need to install 2 `libllama.so`.

Freed-Wu:


abetlen:
@Freed-Wu I'll finish #350 first and add a `NO_BUILD` parameter or something similar

Freed-Wu:
Great. Expect your NO_BUILD. :+1:

PR #511, opened 2023-07-20
Update llama_cpp.py - Fix c_char_p to Array[c_char_p] and c_float to …

viniciusarruda:
…Array[c_float]

### DIFF
--- a/llama_cpp/llama_cpp.py
+++ b/llama_cpp/llama_cpp.py
@@ -748,7 +748,7 @@ def llama_get_vocab(
     return _lib.llama_get_vocab(ctx, strings, scores, capacity)
 
 
-_lib.llama_get_vocab.argtypes = [llama_context_p, c_char_p, c_float, c_int]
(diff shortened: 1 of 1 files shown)


abetlen:
@viniciusarruda had to revert this because Array is not a valid argtype.

viniciusarruda:
Ok. But to run the following code in my machine worked to use Array, otherwise it will raise type error:

```python
lparams = llama_cpp.llama_context_default_params()

ctx = llama_cpp.llama_init_from_file(str(model_path).encode("utf8"), lparams)

n_vocab = llama_cpp.llama_n_vocab(ctx)

strings = (llama_cpp.c_char_p * n_vocab)()
scores = (llama_cpp.c_float * n_vocab)()
n_vocab = llama_cpp.c_int(n_vocab)

llama_cpp.llama_get_vocab(llama_cpp.llama_context_p(ctx), strings, scores, n_vocab)
```

Issue #488, opened 2023-07-18
Llama2


abetlen:
- [x] Test Llama2 models
  - [x] 7b
  - [x] 13b
  - [x] 70b (WIP)
- [ ] Add download links to README

antonkulaga:
I can confirm that Llama 13b model works. However, I am having bizarre errors when I try to use it together with langchain to make embeddings, it gets  inf tokens per second and goes for ages.

abetlen:
@antonkulaga will investigate, does this issue only come up with the embeddings? It could be an upstream issue in llama.cpp

bretello:
Testing out 70b (quantized) on an M1 max with 64GB of RAM:
```console
ins] In [2]: from llama_cpp import Llama

[ins] In [3]: MODEL_PATH = "./llama2/70b-v2-q4_0.bin"

[ins] In [4]: model = Llama(model_path=MODEL_PATH)
llama.cpp: loading model from ./llama2/70b-v2-q4_0.bin
llama_model_load_internal: format     = ggjt v3 (latest)
llama_model_load_internal: n_vocab    = 32000
llama_model_load_internal: n_ctx      = 512
llama_model_load_internal: n_embd     = 8192
llama_model_load_internal: n_mult     = 4096
llama_model_load_internal: n_head     = 64
llama_model_load_internal: n_layer    = 80
llama_model_load_internal: n_rot      = 128
llama_model_load_internal: freq_base  = 10000.0
llama_model_load_internal: freq_scale = 1
llama_model_load_internal: ftype      = 2 (mostly Q4_0)
llama_model_load_internal: n_ff       = 24576
llama_model_load_internal: model size = 65B
llama_model_load_internal: ggml ctx size =    0.19 MB
error loading model: llama.cpp: tensor 'layers.0.attention.wk.weight' has wrong shape; expected  8192 x  8192, got  8192 x  1024
llama_load_model_from_file: failed to load model
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 model = Llama(model_path=MODEL_PATH)

[...]

PR #522, opened 2023-07-24
Llama2 70b support

bretello:
Add support for llama2 70b (uses `n_gqa` parameter)

See https://github.com/abetlen/llama-cpp-python/issues/488#issuecomment-1647728733 and https://github.com/ggerganov/llama.cpp/pull/2276

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,14 @@ For instance, if you want to work with larger contexts, you can expand the conte
 llm = Llama(model_path="./models/7B/ggml-model.bin", n_ctx=2048)
 ```
 
+### Loading llama-2 70b
+
+Llama2 70b must set the `n_gqa` parameter (grouped-query attention factor) to 8 when loading:
+
+```python
+llm = Llama(model_path="./models/7B/ggml-model.bin", n_gqa=8)
+```
+
 ## Web Server
 
 `llama-cpp-python` offers a web server which aims to act as a drop-in replacement for the OpenAI API.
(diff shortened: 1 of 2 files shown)


bretello:
From my limited testing, it seems that the GPU is not being used when providing `n_gpu_layers=1`, even when compiled with metal. I tried running other models (yet) but I'm probably going to work on these in the coming days.

oobabooga:
I have made a test and was able to get 70b working with GPU acceleration by compiling it like this (probably half of it is redundant):

```
pip uninstall llama-cpp-python
git clone 'https://github.com/bretello/llama-cpp-python' -b llama2-70b-support
cd llama-cpp-python/
cd vendor/llama.cpp/
make LLAMA_CUBLAS=1 -j8
cd ..
CMAKE_ARGS="-DLLAMA_CUBLAS=on" FORCE_CMAKE=1 pip install . --no-cache-dir
```

oobabooga:
Maybe `rms_norm_eps` is also necessary?

https://github.com/ggerganov/llama.cpp/pull/2374

It seems like it is currently necessary to manually set it to `1e-5` for llama-2 (the default is `1e-6`).

bretello:
@oobabooga Rebased on `main` (v0.1.76), now that's also included.

abetlen:
@bretello thanks for rebasing, I'll merge this in and publish a new release since I don't think it can work without setting that param during initialisation.

PR #523, opened 2023-07-25
Update tensor_split to match llama.cpp's change

shouyiwang:
[Llama.cpp changed its tensor_split from an array to a pointer](https://github.com/ggerganov/llama.cpp/commit/ae178ab46bfd6ecb2422da5dad441a4e2fef8b7e) 4 days ago, so our implementation doesn't work anymore. 
I updated the code to make it work.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -273,13 +273,12 @@ class Llama:
         self.params.low_vram = low_vram
 
         self.tensor_split = tensor_split
-        self._c_tensor_split = None
+        self._p_tensor_split = None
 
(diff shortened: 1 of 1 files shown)


slaren:
You could look into what parameters are being passed to `llama_eval`, this could be caused by a batch size larger than 512 (`n_tokens`), or a value of `n_past` higher than `n_ctx`. We should probably add some checks.

shouyiwang:
@abetlen 
When will you be releasing a new version? I've noticed that we've been stuck at 0.1.77 for over three weeks now.

PR #565, opened 2023-08-03
add `mul_mat_q` parameter

bretello:
- add `mul_mat_q` parameter
- Update llama.cpp

### DIFF
--- a/llama_cpp/llama_cpp.py
+++ b/llama_cpp/llama_cpp.py
@@ -181,6 +181,7 @@ llama_progress_callback = ctypes.CFUNCTYPE(None, c_float, c_void_p)
 
 #     // Keep the booleans together to avoid misalignment during copy-by-value.
 #     bool low_vram;   // if true, reduce VRAM usage at the cost of performance
+#     bool mul_mat_q;  // if true, use experimental mul_mat_q kernels
 #     bool f16_kv;     // use fp16 for KV cache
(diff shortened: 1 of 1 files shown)


bretello:
Adding some context: trying to load Llama 70b on a M1 Macbook resulted in a segfault.

The first commit in this PR fixes this segfault, possibly caused by the boolean flags not being interpreted correctly due to  `mul_mat_q` missing.

The second commit updates `llama.cpp`. Without this change loading/inference with a 70b model would result in the following assertion failures:
```console
GGML_ASSERT: llama-cpp-python/vendor/llama.cpp/ggml-metal.m:612: ne02 == ne12
GGML_ASSERT:  llama-cpp-python/vendor/llama.cpp/ggml-metal.m:612: ne02 == ne12
GGML_ASSERT:  llama-cpp-python/vendor/llama.cpp/ggml-metal.m:612: ne02 == ne12
```
Note that this only happened with the 70b model

Dampfinchen:
Yes, this is a very important change. Prompt processing speed is a little faster for non-k variants and it saves a lot of VRAM. Right now k-quants are slower but this is going to change soon enough.

hajowieland:
Any ETA when this will be merged? Currently the `llama-2-70b-chat.ggmlv3.q4_0.bin` model can not be used because of the segfaults / assertion failures.
Thanks @bretello for your work!

dillfrescott:
I am still getting this error when attempting to run 70B models. Please help

PR #567, opened 2023-08-04
fix the problem of fasle autoregressive generation

hzg0601:
The very key point of the generative language model is autoregressive, which is to utilize current tokens to regressively get the next token. However, the implementation of current codes is not real autoregression. It only uses two tokens to execute Llama.eval, which would make the generated result completely random. Though faster, it's not the right way how LLM works.
One can check that by printing the number of tokens in Llama.eval
before

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -467,18 +467,18 @@ class Llama:
             if return_code != 0:
                 raise RuntimeError(f"llama_eval returned {return_code}")
             # Save tokens
-            self.input_ids[self.n_tokens : self.n_tokens + n_tokens] = batch
+            self.input_ids[i : i + n_tokens] = batch
             # Save logits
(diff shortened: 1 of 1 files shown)


abetlen:
@hzg0601 you're correct that only the new tokens are passed into eval and this appears to be incorrect however what's actually happening is that past tokens are saved inside the llama.cpp kv cache for the given model. When making subsequent calls to llama_eval you only need to increase the n_past parameter and pass in the new tokens and the model internally usese the past tokens for the autoregressive prediction.

hzg0601:
Sorry for the bothering! I should have learned more deeply about llama-cpp in the first place.

hzg0601:
And I still have a little question about it, I wonder if it's convenient for you to answer it for me: To my best knowledge, if kv cache is adopted, then only one token is needed, why current implementation use two? Is this anther arrangement of llama-cpp too?

PR #575, opened 2023-08-05
Create simple_local_chat.py

Mrgithub93:
easy to use high level chat script

### DIFF
--- a/examples/high_level_api/simple_local_chat.py
+++ b/examples/high_level_api/simple_local_chat.py
@@ -2,7 +2,7 @@ from llama_cpp import ChatCompletionMessage, Llama
 import argparse
 
 parser = argparse.ArgumentParser()
-parser.add_argument("-m", "--model", "--interactive", type=str, default="<your path to your ggml-model.bin>")
+parser.add_argument("-m", "--model", type=str, default="<your path to your ggml-model.bin>")
 args = parser.parse_args()
 
 llm = Llama(model_path=args.model)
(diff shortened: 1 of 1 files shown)


examples/high_level_api/simple_local_chat.py
```diff
@@ -0,0 +1,36 @@
+from llama_cpp import ChatCompletionMessage, Llama
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-m", "--model", "--interactive", type=str, default="<your path to your ggml-model.bin>")
```
bretello, line 5: Why `--interactive`?

examples/high_level_api/simple_local_chat.py
```diff
@@ -0,0 +1,36 @@
...
+args = parser.parse_args()
+
+llm = Llama(model_path=args.model)
+
+# Create a list of messages
+messages = [
+    ChatCompletionMessage(role='system', content='start chat'),
+    ChatCompletionMessage(role='user', content='Hello)
```
bretello, line 13: Missing end single quote `'`

bretello (changes requested):

examples/high_level_api/simple_local_chat.py
Mrgithub93: Damn, forgot that one

examples/high_level_api/simple_local_chat.py
Mrgithub93: Wtf

delock:
Hi, thanks for this example!

When I keep talking with it I got this error, is there a way to avoid this error?
```
ValueError: Requested tokens (527) exceed context window of 512
```

Issue #597, opened 2023-08-10
Cuda Dockerfile - model produces garbage output on GPU (if layers offloaded to GPU)


pradhyumna85:
# Current Behavior

If we use the cuda dockerfile (https://github.com/abetlen/llama-cpp-python/blob/main/docker/cuda_simple/Dockerfile) and use it with any ggml model file and offloading layers to gpu, model outputs garbage text.

By garbage I mean, absolute garbage - symbols, special characters, Unicode characters etc, no English or any language.

# Environment and Context

Running on centos 7 with docker and nvidia docker toolkit installed
Instance - aws g4dn.12xlarge (4 Nvidia T4 - 64GB VRAM)

# Exact same issue seen in a repo related to Cublas in koboldcpp which also kind of uses llamacpp

issue: https://github.com/noneabove1182/koboldcpp-docker/issues/1
fix commit: https://github.com/noneabove1182/koboldcpp-docker/commit/331326a0e340ac845855d346f730356b000650c7

# Fix
I'll raise a pull request, with the updated dockerfile.

Cheers !🥂

afafw:
Experienced similar issue, it returns stuff like content": "The\u0015\u0016\u0013\t\u0018!\u001d\u0019\u000e\u001f\u0005\u001e\u0011 ". Which is obviously too small for any normal utf-8 chars.

afafw:
BTW, you can try this dockerfile. It worked for me.
[Dockerfile.txt](https://github.com/abetlen/llama-cpp-python/files/12342701/Dockerfile.txt)

pradhyumna85:
> Experienced similar issue, it returns stuff like content": "The\u0015\u0016\u0013\t\u0018!\u001d\u0019\u000e\u001f\u0005\u001e\u0011 ". Which is obviously too small for any normal utf-8 chars. 

The only meaningful word in your output - "The" probably came due to not all layers offloaded to GPU(s), am I correct?

afafw:
Actually, that wasn't the case, I loaded all 35 layers into it. But I understand what you mean, since I did tried to load only part of network into gpu and it wasn't completely rubbish. It was just generating weird words.

afafw:
I think the key is also "ENV CUDA_DOCKER_ARCH=all"  as I just saw your pull request.

pradhyumna85:
> BTW, you can try this dockerfile. It worked for me.
[...]

PR #598, opened 2023-08-10
Fixed Cuda Dockerfile

pradhyumna85:
PR for the issue: https://github.com/abetlen/llama-cpp-python/issues/597

Previously models produced garbage output when running on GPU with layers offloaded.

### DIFF
--- a/docker/cuda_simple/Dockerfile
+++ b/docker/cuda_simple/Dockerfile
@@ -4,13 +4,24 @@ FROM nvidia/cuda:${CUDA_IMAGE}
 # We need to set the host to 0.0.0.0 to allow outside access
 ENV HOST 0.0.0.0
 
+RUN apt-get update && apt-get upgrade -y \
+    && apt-get install -y git build-essential \
+    python3 python3-pip gcc wget \
+    ocl-icd-opencl-dev opencl-headers clinfo \
+    libclblast-dev libopenblas-dev \
+    && mkdir -p /etc/OpenCL/vendors && echo "libnvidia-opencl.so.1" > /etc/OpenCL/vendors/nvidia.icd
+
 COPY . .
 
-# Install the package
-RUN apt update && apt install -y python3 python3-pip
+# setting build related env vars
+ENV CUDA_DOCKER_ARCH=all
+ENV LLAMA_CUBLAS=1
+
+# Install depencencies
 RUN python3 -m pip install --upgrade pip pytest cmake scikit-build setuptools fastapi uvicorn sse-starlette pydantic-settings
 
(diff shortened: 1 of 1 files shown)


asteinba:
Hey @pradhyumna85,

I think I just ran into this. What do you mean with garbage? Do you mean it still produces valid sentences but the answers doesn't make sense?

And can you also explain how the fix works/what was the reason :). 

Thanks! Understanding this would help me a lot 😁

pradhyumna85:
Hi @asteinba,
By garbage I mean, absolute garbage - symbols, special characters, Unicode characters etc, no English or any language.

The most relevant change in the Dockerfile is the environment variable - **CUDA_DOCKER_ARCH=all**
It is mentioned in the readme of the official llama.cpp repo (https://github.com/ggerganov/llama.cpp) in the "Docker with CUDA section"

It basically passes the **-Wno-deprecated-gpu-targets** to the **nvccflags** in the llama.cpp **makefile**. (https://github.com/ggerganov/llama.cpp/blob/master/Makefile)
```makefile
.
.
.
ifdef LLAMA_CUBLAS
	CFLAGS    += -DGGML_USE_CUBLAS -I/usr/local/cuda/include -I/opt/cuda/include -I$(CUDA_PATH)/targets/x86_64-linux/include
	CXXFLAGS  += -DGGML_USE_CUBLAS -I/usr/local/cuda/include -I/opt/cuda/include -I$(CUDA_PATH)/targets/x86_64-linux/include
	LDFLAGS   += -lcublas -lculibos -lcudart -lcublasLt -lpthread -ldl -lrt -L/usr/local/cuda/lib64 -L/opt/cuda/lib64 -L$(CUDA_PATH)/targets/x86_64-linux/lib
	OBJS      += ggml-cuda.o
	NVCCFLAGS = --forward-unknown-to-host-compiler -use_fast_math
ifdef LLAMA_CUDA_NVCC
	NVCC = $(LLAMA_CUDA_NVCC)
else
	NVCC = nvcc
endif #LLAMA_CUDA_NVCC
ifdef CUDA_DOCKER_ARCH
	NVCCFLAGS += -Wno-deprecated-gpu-targets -arch=$(CUDA_DOCKER_ARCH)
else
	NVCCFLAGS += -arch=native
endif # CUDA_DOCKER_ARCH
.
.
.
```

**CMAKE_ARGS="-DLLAMA_CUBLAS=on" FORCE_CMAKE=1 pip install llama-cpp-python**, I just copied from the readme (https://github.com/abetlen/llama-cpp-python) instruction for cublas installation.

asteinba:
Thank you very much for the explanation! I really appreciate that :)

abetlen:
@pradhyumna85 thank you for the fix, lgtm

PR #604, opened 2023-08-12
Add doc string for n_gpu_layers argument and make -1 offload all layers

aliencaocao:


gaby:
If it wasnt for this, i wouldnt know you can use `-1`

aliencaocao:
Actually -1 is not tested...it works for you?

gaby:
> Actually -1 is not tested...it works for you?

I haven't tried it

aliencaocao:
Well now it should work

abetlen (approved):

PR #621, opened 2023-08-17
Fix typos in llama_grammar

c0sogi:
It should compare between enum and enum, not enum and enum.value.
This can lead to grammar printing problem.


abetlen:
@c0sogi thank you for the contribution!

PR #678, opened 2023-09-07
Adds Helm-Chart for easy Kubernetes deployment

3deep5me:
Maybe useful for Kubernetes users 😊.


extradosages:
This is a library; it should probably be kept infrastructure agnostic. This is probably more appropriate for a project that _uses_ `llama-cpp-python`.

3deep5me:
This chart can also be used as a library or "sub-chart". That's the way I use it :). But I understand the point.

PR #680, opened 2023-09-07
Fix low level api examples

jbochi:
This fixes the documentation of the low-level API in `README.md` and in the `examples` folder.

`llama_init_from_file`, for example, was replaced by `llama_load_model_from_file` in https://github.com/abetlen/llama-cpp-python/commit/952228407ebd68ef621ad747e3561c821d1c02d3


abetlen:
@jbochi thank you for the contribution!

PR #688, opened 2023-09-11
Update examples from ggml to gguf and add hw-accel note for Web Server

jasonacox:
Minor README documentation updates:

* Convert references in examples from `ggml` models to `gguf`
* Add comment to Web Server section that Hardware Acceleration can be applied to the server installation as well:

> Similar to Hardware Acceleration section above, you can also install with GPU (cuBLAS) support like this:
>
>```bash
> CMAKE_ARGS="-DLLAMA_CUBLAS=on" FORCE_CMAKE=1 pip install llama-cpp-python[server]
>python3 -m llama_cpp.server --model models/7B/gguf-model.bin --n_gpu_layers 35
>```


abetlen:
@jasonacox thank you! I think for the filenames the file extension actually changed from `.bin` to `.gguf` so something like `model.gguf` or similar would be a little better.

jasonacox:
Great idea @abetlen !  I added a commit to change these examples to llama-model.gguf.

PR #694, opened 2023-09-12
CLI chatbot example for llama-cpp-python[server] 

jasonacox:
## CLI ChatBot Example for llama-cpp-python[server] 

Python chat client for OpenAI and the llama-cpp-python[server] OpenAI API Compatible 
Web Server. Provides a simple command line interface (CLI) chat session.

Features:
  * Uses OpenAI API
  * Works with local hosted OpenAI compatible llama-cpp-python[server]
  * Retains conversational context for LLM
  * Uses response stream to render LLM chunks instead of waiting for full response

Requirements:
  * pip install llama-cpp-python[server]
  * python3 -m llama_cpp.server --model models/7B/ggml-model.bin

### DIFF
--- a/examples/server/chatbot.py
+++ b/examples/server/chatbot.py
@@ -1,0 +1,97 @@
+#!/usr/bin/python3
+"""
+Llama-cpp-python[server] - Simple CLI ChatBot Example
+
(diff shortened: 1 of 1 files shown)


earonesty:
nice!

PR #703, opened 2023-09-13
Add X-Request-ID request header for mirroring custom IDs.

### LINKED ISSUE #689: Allow `X-Request-ID` or equivalent in request headers and return in the response.
Currently there is no way of tracking requests for `llama-cpp-server` afaik (I may miss a related feature). It would be really good if we can put custom IDs/tokens for client side that will be returned by the server. That would be very useful for backtracking, handling async requests, etc.

Usually [Request-ID](https://http.dev/x-request-id) or `Correlation-ID`. A [related read](https://stackoverflow.com/q/56068619/7871601) about using which one. This is easy to implement, just returning an id in the response headers; however, it would significantly impact the users imho.

devrimcavusoglu:
Closes #689.

## Example
image


abetlen:
@devrimcavusoglu thank you! Looks good.

Issue #571, opened 2023-08-05
Support for NUMA


rankaiyx:
The numa feature of llama.cpp does not seem to be supported, resulting in significant performance degradation on servers with multiple numa nodes.

Additional Context
NUMA support
--numa: Attempt optimizations that help on some systems with non-uniform memory access. This currently consists of pinning an equal proportion of the threads to the cores on each NUMA node, and disabling prefetch and readahead for mmap. The latter causes mapped pages to be faulted in on first access instead of all at once, and in combination with pinning threads to NUMA nodes, more of the pages end up on the NUMA node where they are used. Note that if the model is already in the system page cache, for example because of a previous run without this option, this will have little effect unless you drop the page cache first. This can be done by rebooting the system or on Linux by writing '3' to '/proc/sys/vm/drop_caches' as root.
https://github.com/ggerganov/llama.cpp/blob/master/examples/main/README.md

This downstream project needs it.
https://github.com/oobabooga/text-generation-webui/issues/3444

abetlen:
@rankaiyx @StoyanStAtanasov just added support for numa in v0.2.4

PR #704, opened 2023-09-13
Add "numa" parameter

StoyanStAtanasov:
Add "numa" parameter ref #571
ref https://github.com/oobabooga/text-generation-webui/issues/3444

used "add mul_mat_q parameter #565" as example


abetlen:
@StoyanStAtanasov just added this in v0.2.4, this has to be done via the `llama_init_backend` function which has to be executed before any other llama.cpp api calls hence the delay on this.

StoyanStAtanasov:
@abetlen Wow, I was way off on this🤦‍♂️. Thanks 🙏. It's a different story when you know what you are doing 👍

PR #684, opened 2023-09-10
PR: [issue: 672] add __version__ 

janvdp:
This a PR for #672 adding the `__version__`. 

This PR includes the features mentioned in the issue, but no changes have been made to support poetry automatically using the llama_cpp/version.py version.

If you merge this PR as is, you still need to update llama_cpp/version.py as well as pyproject.toml.

Feel free to already merge, or provide feedback as to first adding poetry support.

PR #713, opened 2023-09-14
Update pyproject.toml

janvdp:
Additional PR for https://github.com/abetlen/llama-cpp-python/pull/684

Updated pyproject.toml to read the version dynamically following the setuptools docs: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,3 +72,5 @@ Changelog = "https://llama-cpp-python.readthedocs.io/en/latest/changelog/"
 [tool.pytest.ini_options]
 addopts = "--ignore=vendor"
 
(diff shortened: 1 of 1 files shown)


abetlen:
Hey @janvdp from version 0.2 on I started using the dynamic metadata regex tool from https://scikit-build-core.readthedocs.io/en/latest/configuration.html#dynamic-metadata to set the version which is now found in the root `__init__.py` file.

janvdp:
@abetlen Aha, thanks for the info. I did not know. Had a look at the docs, found it a little obscure, but if it works, let's roll. 

Safe to reject/delete this PR.

PR #716, opened 2023-09-14
gguf reader layer and size estimates

earonesty:
i've found that without some sort of layer and size estimate it's very hard to choose the right number of layers to offload

todo:

 - get a size estimate based on needed context size!

### DIFF
--- a/llama_cpp/__init__.py
+++ b/llama_cpp/__init__.py
@@ -1,4 +1,5 @@
 from .llama_cpp import *
 from .llama import *
+from .gguf_reader import *
 
 __version__ = "0.2.4"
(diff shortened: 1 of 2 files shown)


abetlen:
Hey @earonesty this makes sense and I do want to integrate gguf more closely into `llama-cpp-python`. Is it possible to use the pip published `gguf` package to reduce the amount of maintenance required when that's updated?

earonesty:
> Hey @earonesty this makes sense and I do want to integrate gguf more closely into `llama-cpp-python`. Is it possible to use the pip published `gguf` package to reduce the amount of maintenance required when that's updated?

unfortunately that package has no reader support.   i used the source for that to reverse engineer the format and write the reader!

PR #733, opened 2023-09-18
Add cancel() method to interrupt a stream

### LINKED ISSUE #599: Dynamically intterupt token generation
During the generation of tokens I would like to stop when I encounter some condition that changes during the runtime, using `Stream=True`.
E.g. I would like to stop generation after 5 lines of generation.

I would like a method on llm called `stop()`, or `interrupt()`, that forces the model to stop after the next token is generated, similar to CTRL+C in the regular llama.cpp

**Describe alternatives you've considered**
I have considered adding a newline as stop token, but I think this is not performant. Another way I can think of is changing the `stop` list after passing it the generation method, but that feels hacky.

simonchatts:
Fixes #599.

Thanks for all your work on this project!

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -392,6 +392,7 @@ class Llama:
         self.scores: npt.NDArray[np.single] = np.ndarray(
             (n_ctx, self._n_vocab), dtype=np.single
         )
+        self.cancel_stream = False
 
     @property
     def _input_ids(self) -> npt.NDArray[np.intc]:
@@ -973,6 +974,11 @@ class Llama:
             if token == self._token_eos:
                 text = self.detokenize(completion_tokens)
                 finish_reason = "stop"
+                break
+
+            if self.cancel_stream:
+                finish_reason = "cancel"
+                self.cancel_stream = False
                 break
 
             completion_tokens.append(token)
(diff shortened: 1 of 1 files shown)


tk-master:
please accept this pr @abetlen

tk-master:
Actually.. I found an issue with this method.. this will only cancel after a token is generated but if the llm is slow or gets stuck processing the prompt, this doesn't cancel it.. 

We need a better method.

tk-master:
I'm coming back to this because I need to figure out a better method to interrupt the generation programmatically..

For a console-based scenario it's pretty easy in python, all I have to do is surround the code with try except KeyboardInterrupt: .. then I can just press ctrl+c at any point to gracefully interrupt the llm.. 

**But**.. if I'm using a front-end user interface, I haven't managed to make it work _properly_ let's say with a button "Stop generating" that can call a python function.. because of the issue I mentioned in the previous post..

@abetlen sorry to bother again but do you have any suggestions/ideas on how to accomplish this?

woheller69:
Why not add it now and improve if there is a better solution. For now this would work in most cases.

woheller69:
has anyone found a reasonable solution for this? Or am I the only one not willing to wait until the model finishes without killing the job and losing context?

2gog:
Any chance this gets merged for now?

madprops:
It indeed blocks until the first token is produced, but cancelling it after that is trivial. The other similar issue is cancelling a model that is loading.

woheller69:
gpt4all python bindings offer a similar way which allows stopping with the next token

ekcrisp:
+1 can we merge this?

bdashore3:
Take a look at https://github.com/ggerganov/llama.cpp/issues/10509 which should permanently solve this problem on lcpp's side

PR #748, opened 2023-09-24
Adds openai-processing-ms response header

Tradunsky:
Original OpenAI endpoint contains useful header for measuring pure processing time in ms.
зображення

This helps to avoid additional network delay, geo locality overhead time into final measurements.


abetlen:
@Tradunsky lgtm!

PR #772, opened 2023-09-30
fix: recognize `embedding=True` argument

moritzwilksch:
seems like we set the `embedding` attribute in `context_params` here:
https://github.com/abetlen/llama-cpp-python/blob/bca965325d236fe68c64d8f063b1887b111d81d8/llama_cpp/llama.py#L342

but try to access it in `model_params` later which always raises an error.


abetlen:
Hey @moritzwilksch I just saw this after pushing a fix as well, thank you very much though!

PR #784, opened 2023-10-02
Add functionary support

abetlen:
# Overview
Adds support for proper OpenAI function calling using the [functionary](https://github.com/MeetKai/functionary) language model.

This model is capable of performing openai style funciton calling, selecting functions dynamically, and regular chat completions.

[Link to model in GGUF format](https://huggingface.co/abetlen/functionary-7b-v1-GGUF/tree/main)

- [x] Functionary Custom Format
- [x] Functionary Response Formatting (tricky)
- [x] Test with OpenAI tests and instructor library
- [x] Clean up API

# Notes
It was straightforward enough to fit Functionary's prompt formatting into the new format method
image
and the quantized model behaves as expected however it's clear that the current API is inadequate to handle function calling properly
image

To do this correctlly we need to
- Format the incoming message requests to functionary format
- Generate up to  the end of the assistant message (next `:` character) to see if any function is called while not returning any of the generated tokens up until that point.
- If it is perform constrained generation and start returning tokens

# Update: It Works!
image


teleprint-me:
I noticed functionary is specific to a model even though grammar is general enough to support **any** model. Not a big deal as I still haven't had time to look at the updates yet. I'll be doing that tonight and tomorrow. I'll (hopefully) know and understand more about the status quo by then.

Issue #717, opened 2023-09-14
implement prompt template for chat completion


ehartford:
**Is your feature request related to a problem? Please describe.**
When generating chat completion, it is hard-coded to generate a non-standard prompt template that looks something like:

```
### User: <blabla>
### Assistant: <blabla>
```

system message is currently ignored.

https://github.com/abetlen/llama-cpp-python/blob/255d653ae3bd08c690dbf01f9533daf75f71217c/llama_cpp/llama.py#L1578

This mostly works for most models.  But it's not correct.

**Describe the solution you'd like**

1. add a set of built-in prompt templates user can specify at inference time ["vicuna","alpaca","chatml","llama2-chat"] at minimum
2. recommend copying design from ooba's [instruction templates](https://github.com/oobabooga/text-generation-webui/blob/main/instruction-templates/Alpaca.yaml) or fastchat's [conversation](https://github.com/lm-sys/FastChat/blob/3149253988ee16b0945aa0a381a42a07b8a7829e/fastchat/conversation.py#L384)
3. add ability to pass a template string for other nonstandard formats (such as the one currently implemented in llama-cpp-python).

**Describe alternatives you've considered**
modifying llama-cpp-python to hard code it to llama2-chat format, not a great solution.

**Additional context**

ehartford:
This is a hefty task, with architecture / design elements to do it in a clean way.  I am busy to take it on myself right now, but in a couple weeks I can try if nobody else has done it.

abetlen:
Hey @ehartford this is actually something I've had in the backlog and just started last night in #711 

My plan is to have those format identifiers and also provide a generic class (?) that users can extend to provide a custom chat template. The challenge is that it's not just the prompt that has to be modified but also stop sequences, grammar (in the case of open ai style function calling chats), and a few more things I probably haven't thought about but I think this is do-able.

Thank you for the resources btw!

ehartford:
Awesome thanks!  I think it can be done without requiring the user to write any code, using a clever template system, as is implement by ooba and fastchat.

ehartford:
true,
but that would require rewriting chatbot-ui to use /completions instead of /chat/completions (or using a reverse proxy to do that)

ehartford:
then, my solution is that I will make a proxy that receives calls to /chat/completions, and rewrites them into a call into llama-cpp-python's /completions endpoint, in order to inject the proper prompt format.

teleprint-me:
[...]

PR #790, opened 2023-10-03
#717: Add support for Huggingface Autotokenizer

bioshazard:
I am sold on the idea of using an OpenAI style endpoint and love how this project delivers it. All that remains is a transparent chat templating process. This PR delivers universal chat templating via [Huggingface Autotokenizer](https://huggingface.co/docs/transformers/main/chat_templating). Just set `HF_MODEL` before running the server and use `--chat_model autotokenizer`.  

The templates hugging face provides don't quite seem to deliver a perfect outcome, but its got huge potential:

- Nous hermes 13b  doesn't include "### Response" in the prompt so the reply includes it and has to be stripped manually
- mistral 7b instruct errors if you include a system message instead of just dropping it

I plan to use this method to add my own customized templates to hugging face. This solution also adds a default stop token provided by the model. I see this as a clear path to universal chat templating that includes its own stop token so you can swap models while always keeping the same chat payload.

I also really like the use of HF_MODEL with env var as inspired by [12 Factor App](https://12factor.net/config), but any feedback or change is welcome to get this into the official repo. Thanks!

Relates to https://github.com/abetlen/llama-cpp-python/issues/717


abetlen:
@bioshazard great work, very good idea! Give me a few days to settle on some  decisions for this API and I'll merge this in!

teleprint-me:
Hey, would you mind looking at my draft #809? Maybe we can squeeze this in there.

That way we can incorporate the best of both worlds.

bioshazard:
Zephyr decided to have its [own template](https://huggingface.co/TheBloke/zephyr-7B-alpha-GGUF/blob/main/README.md). Exactly the kind of use case I want. @teleprint-me does your PR account for the stop token  when using autotokenizer?

teleprint-me:
@bioshazard 

That's ChatML, so yes. It does account for it. 

It wouldn't matter because I'm attempting to design my PR in a way that let's you define your own if necessary. So, you would just plug it in after creating a new `Llama` instance and then start using `create_chat_completion` after setting the new custom template.

I'm in the middle of 4 or 5 different projects at the moment, not including work, so it's going to take some time for each of them. I'm only one person and can only do so much.

I'll get around to it though. If you have any time to spare, feel free to review what's there, and pitch some ideas. You can pull in my branch and even make some improvements if you see any room for anything.

gardner:
@bioshazard this is great. I just read the transformers [chat_templating](https://huggingface.co/docs/transformers/main/chat_templating) doc. and came to search for it in this repo as I've run into issues with formats in the past.

For me, using the environment variable `HF_MODEL` is a bit awkward.

bioshazard:
@abetlen have you given any further thought to the use of Autotokenizer?

abetlen:
Hey @bioshazard finally had a chance to merge in my functionary changes which required the introduction of a little more complexity to chat handling but it doesn't look to have impacted this PR.

There's a few changes I want to make to this PR but I'll likely merge it first then make those adjustments:
- This should probably be used as a helper function that registers a chat template given a path to a huggingface model
- When users import the llama_cpp library in python they can use this function to register the template directly and using it in subsequent chat_completion calls
- When using the server we can add a HF_MODEL or similar environment variable that registers the handler and uses it directly.

bioshazard:
That's great news! Thanks for the update and yea sounds good! 😁

PR #795, opened 2023-10-05
Add support for batched decoding api

abetlen:
llama.cpp recently moved to a new api which supports batching (both for single sequences with multiple outputs and multiple seperate streams) and streaming support. This new api based on `llama_decode` supercedes the now deprecated `llama_eval` api. This means that the current api should be migrated anyways regardless of the new features but we'll see how easy it is to implement along the way.


flexorRegev:
is there a way to help with this one?

zpzheng:
Is this feature live yet？Why can't I support batch tasks locally？

PR #809, opened 2023-10-10
Draft: Refactor Chat Formatter for Enhanced Flexibility and Extensibility

teleprint-me:
### Draft:
**Refactor Chat Formatter for Enhanced Flexibility and Extensibility**

### Description:

**Summary**:
This PR aims to refactor the chat formatter to provide a more flexible and extensible framework. It introduces a foundational template and protocol that allows for easy addition and management of various chat formatting styles.

**Details**:

1. **Refactored Chat Formatter**:
   - Introduced a base template (`BASE_TEMPLATE`) to serve as the foundation for various chat formatting styles.
   - Implemented a `ChatFormatterTemplate` protocol, which provides a standardized interface for all formatter classes.
   - Extracted commonly used methods like `_get_system_message` and `_map_roles` to this protocol, promoting code reuse and reducing redundancy.
   
2. **Llama2 Formatter**:
   - Derived a specific formatter, `Llama2Formatter`, from the `ChatFormatterTemplate`.
   - Specifically handles the "Llama2" chat format.
   - Addressed and fixed a bug related to message formatting.
   
3. **Testing and Validation**:
   - Introduced a sequence of test messages for validation.
   - Created a pytest test to validate the formatter's functionality.
   - Incorporated a pytest fixture for cleaner and more reusable test setup.
   
4. **Commit Strategy**:
   - Organized the changes into descriptive commits with clear naming.

5. **Future Goals**:
   - Enhance the codebase's flexibility, allowing for easy creation and integration of new templates for different models.
   - Empower users to introduce their templates if a suitable one doesn't exist, furthering the codebase's extensibility and user-friendliness.

---

**Feedback Requested**:
I'm looking for early feedback on the current approach, especially concerning the overall structure and the introduction of the new `ChatFormatterTemplate` protocol. Any suggestions or improvements are welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.2.10]
+
+- Fix bug 'llama_model_params' object has no attribute 'embedding' by @abetlen in 42bb721d64d744242f9f980f2b89d5a6e335b5e4
+
 ## [0.2.9]
 
 - Fix critical bug in pip installation of v0.2.8 due to `.git` directory in ac853e01e1a217a578080a4e1b851d2d08450adf
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.2.9]
+
+- Fix critical bug in pip installation of v0.2.8 due to `.git` directory in ac853e01e1a217a578080a4e1b851d2d08450adf
+
 ## [0.2.8]
 
 - Update llama.cpp to ggerganov/llama.cpp@40e07a60f9ce06e79f3ccd4c903eba300fb31b5e
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.2.11]
+
+- Fix bug in `llama_model_params` object has no attribute `logits_all` by @abetlen in d696251fbe40015e8616ea7a7d7ad5257fd1b896 
+
 ## [0.2.10]
 
 - Fix bug 'llama_model_params' object has no attribute 'embedding' by @abetlen in 42bb721d64d744242f9f980f2b89d5a6e335b5e4
--- a/docs/install/macos.md
+++ b/docs/install/macos.md
@@ -38,19 +38,19 @@ llama-cpp-python         0.1.68
 
 ```
 
-**(5) Download a v3 ggml model**
- - **ggmlv3**
- - file name ends with **q4_0.bin** - indicating it is 4bit quantized, with quantisation method 0
+**(5) Download a v3 gguf v2 model**
+ - **ggufv2**
+ - file name ends with **Q4_0.gguf** - indicating it is 4bit quantized, with quantisation method 0
 
-https://huggingface.co/TheBloke/open-llama-7b-open-instruct-GGML
+https://huggingface.co/TheBloke/CodeLlama-7B-GGUF
 
 
 **(6) run the llama-cpp-python API server with MacOS Metal GPU support**
 ```
 # config your ggml model path
-# make sure it is ggml v3
+# make sure it is gguf v2
 # make sure it is q4_0
-export MODEL=[path to your llama.cpp ggml models]]/[ggml-model-name]]q4_0.bin
+export MODEL=[path to your llama.cpp ggml models]]/[ggml-model-name]]Q4_0.gguf
 python3 -m llama_cpp.server --model $MODEL  --n_gpu_layers 1
 ```
 
--- a/llama_cpp/__init__.py
+++ b/llama_cpp/__init__.py
@@ -1,4 +1,4 @@
 from .llama_cpp import *
 from .llama import *
 
-__version__ = "0.2.10"
+__version__ = "0.2.11"
--- a/llama_cpp/__init__.py
+++ b/llama_cpp/__init__.py
@@ -1,4 +1,4 @@
 from .llama_cpp import *
 from .llama import *
 
-__version__ = "0.2.9"
+__version__ = "0.2.10"
--- a/llama_cpp/__init__.py
+++ b/llama_cpp/__init__.py
@@ -1,4 +1,4 @@
 from .llama_cpp import *
 from .llama import *
 
-__version__ = "0.2.8"
+__version__ = "0.2.9"
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -839,9 +839,10 @@ class Llama:
             An embedding object.
         """
         assert self.ctx is not None
+        assert self.model is not None
         model_name: str = model if model is not None else self.model_path
 
-        if self.model_params.embedding == False:
+        if self.context_params.embedding == False:
             raise RuntimeError(
                 "Llama model must be created with embedding=True to call this method"
             )
@@ -863,7 +864,7 @@ class Llama:
             n_tokens = len(tokens)
             total_tokens += n_tokens
             embedding = llama_cpp.llama_get_embeddings(self.ctx)[
-                : llama_cpp.llama_n_embd(self.ctx)
+                : llama_cpp.llama_n_embd(self.model)
             ]
 
             data.append(
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -439,7 +439,7 @@ class Llama:
     def eval_logits(self) -> Deque[List[float]]:
         return deque(
             self.scores[: self.n_tokens, :].tolist(),
-            maxlen=self._n_ctx if self.model_params.logits_all else 1,
+            maxlen=self._n_ctx if self.context_params.logits_all else 1,
         )
 
     def tokenize(self, text: bytes, add_bos: bool = True) -> List[int]:
@@ -964,7 +964,7 @@ class Llama:
         else:
             stop_sequences = []
 
-        if logprobs is not None and self.model_params.logits_all is False:
+        if logprobs is not None and self.context_params.logits_all is False:
             raise ValueError(
                 "logprobs is not supported for models created with logits_all=False"
             )
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -1625,6 +1625,8 @@ class Llama:
         format = llama_chat_format.get_chat_format(self.chat_format)
         result = format(
             messages=messages,
+            functions=functions,
+            function_call=function_call,
         )
         prompt = result.prompt
         if result.stop is not None:
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -91,6 +91,19 @@ def _format_add_colon_space_single(
     return ret
 
 
+def _format_chatml(
+    system_message: str, messages: List[Tuple[str, Optional[str]]], sep: str
+) -> str:
+    """Format the prompt with the chatml style."""
+    ret = "" if system_message == "" else system_message + sep + "\n"
+    for role, message in messages:
+        if message:
+            ret += role + "\n" + message + sep + "\n"
+        else:
+            ret += role + "\n"
+    return ret
+
+
 @dataclasses.dataclass
 class ChatFormatterResponse:
     prompt: str
@@ -290,3 +303,20 @@ def format_open_orca(
     _messages.append((roles[1], None))
     _prompt = _format_add_colon_space_single(system_message, _messages, sep)
     return ChatFormatterResponse(prompt=_prompt, stop=stop_str)
+
+
+@register_chat_format("chatml")
+def format_chatml(
+    messages: List[llama_types.ChatCompletionRequestMessage],
+    **kwargs: Any,
+) -> ChatFormatterResponse:
+    system_template = """<|im_start|>system
+{system_message}"""
+    system_message = _get_system_message(messages)
+    system_message = system_template.format(system_message=system_message)
+    _roles = dict(user="<|im_start|>user", assistant="<|im_start|>assistant")
+    _sep = "<|im_end|>"
+    _messages = _map_roles(messages, _roles)
+    _messages.append((_roles["assistant"], None))
+    _prompt = _format_chatml(system_message, _messages, _sep)
+    return ChatFormatterResponse(prompt=_prompt)
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -7,6 +7,54 @@ To extend or customize, simply inherit from the ChatFormatter class and override
 
 NOTE: The system message is always assumed to be the first element in a sequence.
 
+NOTE: Users should avoid tampering with special tokens to prevent model issues.
+
+---
+
+# IMPORTANT NOTES:
+
+- The use of the merge operator (|) for dictionaries requires Python 3.9 or higher. Keep in mind that llama-cpp-python supports Python 3.8 and later versions. If you are working with an earlier Python version, consider alternatives such as `dict.update()` or creating a custom function to merge dictionaries. For Python 3.9 or higher, the merge operator simplifies dictionary merging.
+Source: https://docs.python.org/3/library/stdtypes.html?highlight=dict#dict
+
+- Special tokens are crucial for the model's underlying operations, impacting pre-training, fine-tuning, and low-level inference processes. Users should avoid modifying special tokens to prevent issues in the model's output during inference. These issues may manifest as token fixation, repetitive language patterns, contextual derailment, and hallucinations. Improper use of separators and templates can exacerbate these problems.
+
+Example using the llama-2 model and its templating schema:
+
+#  1  <<SYS>>My name is Llama and I am a helpful assistant.<</SYS>>$
+#  2  [INST] Hello Llama, my name is User. What's your name? [/INST]$
+#  3  Hello User, my name is Llama. Nice to meet you!$
+#  4  [INST] What can you do? [/INST]$
+#  5  I can assist you with various tasks, including providing structured output for certain queries.$
+#  6  [INST] How can you assist me in my programming projects? [/INST]$
+#  7  $
+
+This initial example is a proper template format that the model understands. It results in proper output and does not confuse the model.
+
+#  1  <<SYS>>My name is Llama and I am a helpful assistant.<</SYS>>$
+#  2  <s>[INST] Hello Llama, my name is User. What's your name? [/INST]$
+#  3  Hello User, my name is Llama. Nice to meet you!</s>$
+#  4  <s>[INST] What can you do? [/INST]$
+#  5  I can assist you with various tasks, including providing structured output for certain queries.</s>$
+#  6  <s>[INST] How can you assist me in my programming projects? [/INST]$
+#  7  $
+
+This example includes the use of special tokens, and the model may or may not use these tokens as a result. The model is not expecting them during inference, which causes unexpected behavior.
+
+#  1  <<SYS>>My name is Llama and I am a helpful assistant.<</SYS>>$
+#  2  $
+#  3  <s>[INST] Hello Llama, my name is User. What's your name? [/INST]$
+#  4  Hello User, my name is Llama. Nice to meet you!</s>$
+#  5  $
+#  6  <s>[INST] What can you do? [/INST]$
+#  7  I can assist you with various tasks, including providing structured output for certain queries.</s>$
+#  8  $
+#  9  <s>[INST] How can you assist me in my programming projects? [/INST]$
+# 10  $
+
+This example is improperly formatted and causes the model to become confused. The model begins to fixate on tokens, uses language repetition, and eventually derails.
+
+---
+
 # Usage example:
 # Registering a custom formatter
 @ChatFormatterFactory.register_predefined_model("llama-2")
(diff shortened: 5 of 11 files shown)


teleprint-me:
For clarity and simplicity, here are the proposed changes.

```python
import dataclasses
from typing import Dict, List, Optional

from transformers import AutoTokenizer

from . import llama_types

# NOTE: Custom Templates use Jinja2.
# If no template is given, then should default to hf's tokenizer template.
# We can define the model and template on a model-to-model basis,
# however, this should be allowed to be overridden for flexibility and extensibility.
# We only need 2 keys, the model name and the jinja2 template.
#
#   template = {"model": "meta-llama/Llama-2-7b-chat-hf", "template": None}
#
#   or
#
# chat_template = {
#     "model": "meta-llama/Llama-2-7b-chat-hf",
#     "jinja": "{% for message in messages %}{% if message['role'] == 'user' %}{{ bos_token + '[INST] ' + message['content'].strip() + ' [/INST]' }}{% elif message['role'] == 'system' %}{{ '<<SYS>>\\n' + message['content'].strip() + '\\n<</SYS>>\\n\\n' }}{% elif message['role'] == 'assistant' %}{{ '[ASST] '  + message ['content'] + ' [/ASST]' + eos_token }}{% endif %}{% endfor %}",
# }
#
# We can probably employ some kind of method for reading a template it in from a file in necessary.
#
# We leave template empty here because HuggingFace defined it already.
#
# Source: https://huggingface.co/docs/transformers/main/chat_templating
#
# Special Thanks and Credit goes to bioshazard for the idea and preliminary implementation.
# Source: https://github.com/abetlen/llama-cpp-python/pull/790


# NOTE: We can still use this for reverse compatibility with the currently employed API.
# This can be modified, if needed, in the future.
@dataclasses.dataclass
class ChatFormatterResponse:
    prompt: str
    stop: Optional[List[str]] = None


class TokenizerCache:
    _cache: Dict[str, AutoTokenizer] = {}

    @classmethod
    def get_tokenizer(cls, model_name: str) -> AutoTokenizer:
        if model_name not in cls._cache:
            cls._cache[model_name] = AutoTokenizer.from_pretrained(model_name)
        return cls._cache[model_name]


class ChatFormatterTemplate:
    def __init__(self, template: Optional[Dict[str, str]] = None):
        if template:
            self.template = template
        else:
            self.template = {
                "model": "meta-llama/Llama-2-7b-chat-hf",
                "jinja": None,
                "tokenize": False,
            }
        self.tokenizer = TokenizerCache.get_tokenizer(self.template["model"])

    def _format_messages(self, messages: List[Dict[str, str]]) -> str:
        # If a custom template is provided, override the tokenizer's default template
        if self.template.get("jinja"):
            self.tokenizer.chat_template = self.template["jinja"]

        return self.tokenizer.apply_chat_template(
            messages, tokenize=self.template["tokenize"]
        )

    def parse_response(self, messages: List[Dict[str, str]]) -> ChatFormatterResponse:
        formatted_content = self._format_messages(messages)
        return ChatFormatterResponse(
            prompt=formatted_content, stop=[self.tokenizer.eos_token]
        )


class ChatFormatter:
    _chat_formatters: Dict[str, ChatFormatterTemplate] = {}

    def register_chat_format(
        self, model_name: str, template: Optional[Dict[str, str]] = None
    ):
        self._chat_formatters[model_name] = ChatFormatterTemplate(template)

    def get_chat_format(self, model_name: str) -> ChatFormatterTemplate:
        if model_name not in self._chat_formatters:
            raise ValueError(f"Model {model_name} is not registered.")

        return self._chat_formatters[model_name]

    def format(self, model_name: str, messages: List[Dict[str, str]]) -> str:
        formatter = self.get_chat_format(model_name)
        return formatter._format_messages(messages)

    def parse(
        self, model_name: str, messages: List[Dict[str, str]]
    ) -> ChatFormatterResponse:
        formatter = self.get_chat_format(model_name)
        return formatter.parse_response(messages)


# NOTE: Template registration is currently a WIP (work in progress)
```

With this refactoring:

- We have a `TokenizerCache` class which helps in efficient reuse of tokenizers without the overhead of recreating them each time.
- We've got a generic `ChatFormatterTemplate` class that leverages `AutoTokenizer` and can be easily customized with different model names and templates.
- The central `ChatFormatter` class manages different chat formatters and offers methods to register, retrieve, format, and parse using these formatters.

You can easily extend the functionality in the future if needed. However, as of now, this design offers a good balance of simplicity, flexibility, and efficiency.

@bioshazard @abetlen @delock @earonesty 

Let me know what you guys think.

bioshazard:
Great work! FWIW @abetlen I will not be upset if my #790 PR gets canned and this chosen as the preferred implementation. Looking forward to AutoTokenizer support either way!

teleprint-me:
It should be noted that this seriously has it's own set of drawbacks. The previous implementation is way more flexible. I'm thinking we should have a middle ground for several reasons.

- HuggingFace will require several dependencies: `torch`, `sentencepiece`, and `protobuf`. Which version of `torch` will be defaulted to? I suspect most users (including myself) are using `cpu` as a device.
- Not all tokenziers will behave as expected: e.g., where do we get a valid, trusted, tokenizer for fine-tuned gated models? Not all models provide a tokenizer and most GGUF models bake-in the tokenizer as well. I haven't dug too much into the core source code for llama.cpp, but it's getting to the point where I'll need to do more than a cursory code review so I have a deeper understanding.
- This creates a serious level of rigidity by requiring the tokenizer to be downloaded as a result; e.g., HuggingFace wraps a lot of stuff and makes a ton of assumptions taking freedom of choice and flexibility away from the user in exchange for a level of flexibility. This creates presumptions and locks users out as a result.

I feel these points should be considered with a serious level of thought while considering the gravity of the implications of such changes.

The middle ground I'm thinking of might be to scrap the current design, including the one's I've come up with and go for a Jinja2-like template system. Something like what HuggingFace has with their Chat Template system.

It'll be easier to test, implement, and streamline as a result. It'll also be interoperable with the HuggingFace API as well by simply hot-swapping templates if a user wants to. We should also make HuggingFace an optional dependency as well do the previously mentioned concerns I have with this approach.

bioshazard:
@teleprint-me 

- I expect users to create a fresh `venv` for a given `llama-cpp-python[server]` instance. Maybe add a `llama-cpp-python[server-hfat]` or something to manage the optional deps.
- Jinja is the ideal templating scheme in all cases, yes please
- Even if the existing HF models don't all have the support for this yet, I figure this project is a big enough deal that its integration would imply incentive to conform to start being better about including it. As well, one can submit their own chat templates.

teleprint-me:
@bioshazard

Enabling the ability to create custom datasets and fine-tuned models is a valuable aspect for me. Flexibility to deviate from existing templates should be a core principle in the design we settle on. Ultimately, the direction we take largely depends on @abetlen since it's his repository.

I believe it's important for us to collaborate and reach a consensus while striving for maximum flexibility with minimal lines of code (SLoC) and abstraction. Excessive SLoC and abstraction can be a warning sign that we might have done something erroneous in our design.

Regarding the use of Jinja-2, I have reservations mainly because I've been using it for a long time, and its original design wasn't intended for this specific use-case. While it might work as a clever workaround, it also raises concerns about our ability to troubleshoot any potential issues that might arise. To ensure the integrity of the expected output, I strongly recommend implementing thorough testing for this approach and treating it as experimental, given its unconventional usage.

My motivation for advocating certain tools and approaches is rooted in a desire to maintain a proactive, inclusive, and open approach to accommodating other developers preferences. While I do lean towards pragmatism and minimalism, I'm open to considering alternative approaches because my primary goal is to enhance this library and its utility.

I appreciate your support for Jinja-2, even though I express caution in its use.

Please let me know if you have any other thoughts or considerations regarding any of the points I've made. I'm open to feedback and healthy criticism.

teleprint-me:
Another possible approach would be the following:

```python
class ChatFormatterInterface(Protocol):
    def __init__(self, template: Optional[Dict[str, Any]] = None):
        raise NotImplementedError

    def format_messages(
        self, messages: List[llama_types.ChatCompletionRequestMessage]
    ) -> str:
        raise NotImplementedError

    def parse_response(
        self,
        messages: List[llama_types.ChatCompletionRequestMessage],
        **kwargs,
    ) -> ChatFormatterResponse:
        raise NotImplementedError


class AutoTokenizerFormatter(ChatFormatterInterface):
    def __init__(self, template: Optional[Dict[str, str]] = None):
        if template:
            self.template = template
        else:
            self.template = {
                "model": "meta-llama/Llama-2-7b-chat-hf",
                "jinja": None,
                "tokenize": False,
            }
        self.tokenizer = TokenizerCache.get_tokenizer(self.template["model"])

    def format_messages(
        self, messages: List[llama_types.ChatCompletionRequestMessage]
    ) -> str:
        # If a custom template is provided, override the tokenizer's default template
        if self.template.get("jinja"):
            self.tokenizer.chat_template = self.template["jinja"]

        return self.tokenizer.apply_chat_template(
            messages, tokenize=self.template["tokenize"]
        )

    def parse_response(
        self,
        messages: List[llama_types.ChatCompletionRequestMessage],
        **kwargs,
    ) -> ChatFormatterResponse:
        formatted_content = self.format_messages(messages)
        return ChatFormatterResponse(
            prompt=formatted_content, stop=[self.tokenizer.eos_token]
        )


class ChatFormatter:
    _chat_formatters: Dict[str, ChatFormatterInterface] = {}

    @staticmethod
    def register_chat_format(name: str):
        def decorator(cls: Type[ChatFormatterInterface]):
            ChatFormatter._chat_formatters[name] = cls()
            return cls

        return decorator

    def get_chat_format(self, name: str) -> ChatFormatterInterface:
        try:
            return self._chat_formatters[name]
        except KeyError:
            raise ValueError(
                f"Invalid chat format: {name}. Valid formats: {list(self._chat_formatters.keys())}"
            )

    def format(self, name: str, messages: List[Dict[str, str]]) -> str:
        formatter = self.get_chat_format(name)
        return formatter.format_messages(messages)

    def parse(
        self,
        name: str,
        messages: List[llama_types.ChatCompletionRequestMessage],
        **kwargs,
    ) -> ChatFormatterResponse:
        formatter = self.get_chat_format(name)
        return formatter.parse_response(messages)
```

This would create a form of backwards compatability.

```python
# NOTE: Template registration is currently a WIP (work in progress)
# External developers can now use the `@ChatFormatter.register_chat_format`
# decorator to register their own custom formatters.
@ChatFormatter.register_chat_format("llama-2")
class Llama2Formatter(AutoTokenizerFormatter):
    def __init__(self, template: Optional[Dict[str, Any]] = None):
        super().__init__(template)
```

It also allows us to override the implementation if necessary.

```python
@ChatFormatter.register_chat_format("llama-2")
class Llama2Formatter(ChatFormatterTemplate):
    _system_template = "[INST] <<SYS>>\n{system_message}\n<</SYS>>\n\n"
    _roles = dict(user="[INST]", assistant="[/INST]")
    _sep = "\n\n"

    def _get_system_message(self, messages: List[llama_types.ChatCompletionRequestMessage]) -> str:
        try:
            if messages[0]["role"] == "system":
                return self._system_template.format(system_message=messages[0]["content"])
            return ""
        except (IndexError, KeyError):
            return ""

    def _map_roles(self, messages: List[llama_types.ChatCompletionRequestMessage]) -> List[Tuple[str, str]]:
        mapped_messages = []
        for message in messages:
            if message["role"] in self._roles:
                mapped_messages.append((self._roles[message["role"]], message["content"]))
        mapped_messages.append((self._roles["assistant"], None))
        return mapped_messages

    def format_messages(self, messages: List[llama_types.ChatCompletionRequestMessage]) -> str:
        system_message = self._get_system_message(messages)
        mapped_messages = self._map_roles(messages)
        return system_message + self._sep.join([msg for role, msg in mapped_messages if msg])

    def parse_response(self, messages: List[llama_types.ChatCompletionRequestMessage]) -> ChatFormatterResponse:
        formatted_content = self.format_messages(messages)
        return ChatFormatterResponse(prompt=formatted_content)
```

This still requires a more well thought out design though.

tranhoangnguyen03:
I notice that in HF, i can use `tokenizer.apply_chat_template` to transform OpenAI's {'role':...,'content'...} chat template  to the named model's template. Will this come with the the same method?

teleprint-me:
@tranhoangnguyen03 

It would be wrapped. 

We would ideally just pass the template in during instantiation.

I'm still working out the details in the current draft, but I'm considering replacing the current templates with jinja2 templates at some point in the future. 

The caveat between the 2 is that,

- A dictionary based approach is easier to pick up and apply.
- Jinja2 has a learning curve even though it would create a level of interoperability.

I think it would be preferable to manage a single approach while accommodating each interface appropriately.

bioshazard:
> I notice that in HF, i can use `tokenizer.apply_chat_template` to transform OpenAI's {'role':...,'content'...} chat template to the named model's template. Will this come with the the same method?

That was the motivation behind my PR ( https://github.com/abetlen/llama-cpp-python/pull/790 ), tho this appears to be a more comprehensive attempt.

earonesty:
can we just stick with Jinja?

teleprint-me:
@earonesty 

Yeah, I included it in the original draft. 

I'm still waiting on feedback. Also, I have a lot going on and my mental bandwidth is low at the moment.

Jinja2 would already be included as a dependency with transformers, so I don't really see that as an issue.

earonesty:
i was thinking of a lot less structure.   literally just copying the jinja templates from huggingface and executing them in the right context.

for example, this is the mistral template:

```
template = {
 "bos_token": "<s>",
 "chat_template": "{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if message['role'] == 'user' %}{{ '[INST] ' + message['content'] + ' [/INST]' }}{% elif message['role'] == 'assistant' %}{{ message['content'] + eos_token + ' ' }}{% else %}{{ raise_exception('Only user and assistant roles are supported!') }}{% endif %}{% endfor %}",
 "eos_token": "</s>",
}

template = jinja2.Template(template.pop("chat_template")
template.render(context, messages=messages)

```

teleprint-me:
@earonesty 

I think it's crucial to consider what's happening here and what the goal of this draft is.

The ChatML structure is passed in as input, which is a sequence of dictionaries representing roles and their associated content.

That sequence is passed in as input and then reformatted into a structured string that llama.cpp expects as input. **Appropriately formatting that input is the goal here**.

***The problem is the variety and variability of templates that require management***.

This is why @abetlen used the following structure,

```py
@dataclasses.dataclass
class ChatFormatterResponse:
    prompt: str
    stop: Optional[Union[str, List[str]]] = None
```

And returned it as a result.

```py
@register_chat_format("llama-2")
def format_llama2(
    messages: List[llama_types.ChatCompletionRequestMessage],
    **kwargs: Any,
) -> ChatFormatterResponse:
    # ...
    return ChatFormatterResponse(prompt=_prompt)
```

So, if we're to consider your proposal, that would mean having a template string to be parsed for each model baked in, which is admittedly a bit of an improvement, but fails to solve the overarching problem at hand.

Consider your example:

```py
mistral = {
 "bos_token": "<s>",
 "chat_template": "{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if message['role'] == 'user' %}{{ '[INST] ' + message['content'] + ' [/INST]' }}{% elif message['role'] == 'assistant' %}{{ message['content'] + eos_token + ' ' }}{% else %}{{ raise_exception('Only user and assistant roles are supported!') }}{% endif %}{% endfor %}",
 "eos_token": "</s>",
}
```

We technically don't need the special tokens at all because everything necessary to format the string should be included already.

```py
mistral_template = "{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if message['role'] == 'user' %}{{ '[INST] ' + message['content'] + ' [/INST]' }}{% elif message['role'] == 'assistant' %}{{ message['content'] + eos_token + ' ' }}{% else %}{{ raise_exception('Only user and assistant roles are supported!') }}{% endif %}{% endfor %}"
```

This is all we would technically need.

llama.cpp doesn't expect much and there are experimental mechanisms for defining the prefix and postfix, but none of these solutions really solve the root problem of the inherit variability involved in a chat template.

For example, we would need to rely on the user being able to define a template for FIM (Fill-in-the-Middle) models.

This is why I considered @bioshazard original proposal, because `AutoTokenizer` already handles all of the grunt work for us.

I mean, if we really want to reduce the amount of SLoC, we could consider scrapping the joint interfaces and simply go for integrating HuggingFace.

```py
class AutoTokenizerFormatter(ChatFormatterInterface):
    def __init__(self, template: Optional[Dict[str, str]] = None):
        self.template = template or huggingface_template
        self.huggingface_login()
        self.tokenizer = TokenizerCache.get_tokenizer(self.template["model"])

    def __call__(
        self,
        messages: List[llama_types.ChatCompletionRequestMessage],
        **kwargs,
    ) -> ChatFormatterResponse:
        formatted_content = self.format_messages(messages)
        return ChatFormatterResponse(
            prompt=formatted_content, stop=[self.tokenizer.eos_token]
        )

    def huggingface_login(self) -> None:
        token = os.getenv("HF_TOKEN")
        if token is None:
            raise AttributeError(
                "Failed to login to huggingface. "
                "Did you forget to set the `HF_TOKEN` environment variable with your huggingface token?"
            )
        huggingface_hub.login(token)

    def format_messages(
        self, messages: List[llama_types.ChatCompletionRequestMessage]
    ) -> str:
        # If a custom template is provided, override the tokenizer's default template
        if self.template.get("jinja"):
            self.tokenizer.chat_template = self.template["jinja"]

        return self.tokenizer.apply_chat_template(
            messages, tokenize=self.template.get("tokenize", False)
        )
```

But as [I previously stated](https://github.com/abetlen/llama-cpp-python/pull/809#issuecomment-1756006112), this seriously has its own set of drawbacks.

My rationale behind all of this is that we would need to be prepared to parse and build the template, then format it to a string, and then pass it back, even if we chose to use only jinja2. The upside to this approach is it would apparently make it compatible with HuggingFace `AutoTokenizer`, but we wouldn't have a bridge for it anymore.

This, unfortunately, is not really a simple problem. It's one of those deceptively simple problems up until you start trying to solve it.

bioshazard:
I still think hugging face Auto tokenizer is the ultimate solution. Even if you are dissatisfied with the models template you can upload your own for yourself and others to use. And it's introduction into llama CPP py will introduce a clear incentive to have functioning templates. It might not account for all use cases today but seems like the ideal foundation to ultimately incentivize and realize universal transparent chat templating. How many argument I really see against it is immediate satisfaction and the option not to have to download from a hugging face though at that point you could just allow reference to a local json definition as is being provided by hugging-face. This is not to say any of your work is unnecessary but as an appeal to the ultimate solution I still find the auto tokenizer approach highly compelling.

bioshazard:
Seems like most of the drawbacks were the inappropriate excessive dependencies. If auto tokenizers chat templating function could be broken into its own library, would you support its candidacy for universal chat template transparency? Especially if you could refer to a local file too rather than a relative model path.

teleprint-me:
@bioshazard

The problem with HuggingFace is introducing unnecessary dependencies that will literally only be used to gain access to the templates. It's hard to justify this argument in my own mind.

bioshazard:
I deleted my last comment after rereading yours. Valid points about the dependencies. Curious of your thoughts on my replacement reply.

teleprint-me:
@bioshazard

I don't know. I can't really find any valid arguments against @earonesty proposal. I'm only attempting to elucidate the complexity involved in his proposal. Not shut it down. I want him to genuinely think about it. I find this to be a hard problem. I'm hoping we can come up with an elegant solution together.

bioshazard:
Not suggesting his proposal is inappropriate. Maybe I'm just thinking out loud about attempting to write a minimal library to break out the model template functionality as I offered.

earonesty:
I feel like the base idea of just supporting raw jinja , and having it be
compatible with the templates that people typically use in HF is fairly
simple to implement

I'm still okay with leaving the existing Python function based templates as
an option

those are more flexible and maybe there's some things you can do with them
that you can't do with jina

teleprint-me:
@earonesty

I hear you and I'm inclined to feel the same. There's pros and cons to every approach here. It's important to realize that we'll always be making a compromise.

I'm hoping to go for one of those solutions that's only obvious in hindsight where we're all happy with it and it's easy to read, understand, and implement.

If you have something more concrete, rather than an abstract notion, I'm all ears. A concrete snippet exemplifying what you're envisioning would be helpful.

Sharing is caring 😉

teleprint-me:
I made a new draft in a separate branch as this branch now conflicts with recent changes.

[teleprint-me:llama-cpp-python:jinja2-templates](https://github.com/abetlen/llama-cpp-python/compare/main...teleprint-me:llama-cpp-python:jinja2-templates)

It's the same general idea and reduces complexity, sloc, and uses the idea of jinja2 templating.

```py
class AutoChatFormatter(ChatFormatterInterface):
    def __init__(
        self,
        template: Optional[str] = None,
        template_class: Optional[Template] = None,
    ):
        if template is not None:
            self._template = template
        else:
            self._template = llama2_template  # default template

        self._renderer = jinja2.Environment(
            loader=jinja2.BaseLoader(),
            trim_blocks=True,
            lstrip_blocks=True,
        ).from_string(
            self._template,
            template_class=template_class,
        )

    def __call__(
        self,
        messages: List[Dict[str, str]],
        **kwargs: Any,
    ) -> ChatFormatterResponse:
        formatted_sequence = self._renderer.render(messages=messages, **kwargs)
        return ChatFormatterResponse(prompt=formatted_sequence)

    @property
    def template(self) -> str:
        return self._template
```

PR #834, opened 2023-10-22
Do not set `grammar` to `None` for new `LlamaGrammar` objects

mthuurne:
The `LlamaGrammar.grammar` attribute is written by `init()`, but that method always returns `None`, so `__init__()` would then discard the previously written object.

Note that `Llama.generate()` will call `LlamaGrammar.reset()` before using the grammar, so it will also work fine without this change, but `Llama.sample()` does not call the `reset()` method.


abetlen:
@mthuurne thank you!

Issue #838, opened 2023-10-23
Error with special tokens tokenization


earzamastsev:
I try using OpenAI-like API with vicuna LLM: python3 -m llama_cpp.server --n_gpu_layers 43 --model ./models/vicuna-13b-v1.5.Q8_0.gguf --port 8010 --host 0.0.0.0 --chat_format vicuna

Send request to endpoint /v1/chat/completions:
{
"max_tokens": 1024,
"temperature": 0.1,
"messages": [
{
"content": "Hello, what is your name?",
"role": "user"
},
{
"content": "My name is AI-asisstant",
"role": "assistant"
},
{
"content": "Can you repeat your name please?",
"role": "user"
}
]
}

And checking final prompt and final tokens. So, i see this:

PROMPT: A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful, detailed, and polite answers to the user's questions. USER: Hello, what is your name? ASSISTANT: My name is AI-asisstant `</s>` USER: Can you repeat your name please? ASSISTANT:

PROMPT TOKENS: [1, 319, 13563, 1546, 263, 12758, 1404, 322, 385, 23116, 21082, 20255, 29889, 450, 20255, 4076, 8444, 29892, 13173, 29892, 322, 1248, 568, 6089, 304, 278, 1404, 29915, 29879, 5155, 29889, 3148, 1001, 29901, 15043, 29892, 825, 338, 596, 1024, 29973, 319, 1799, 9047, 13566, 29901, 1619, 1024, 338, 319, 29902, 29899, 25101, 303, 424, 829, 29879, 29958, 11889, 29901, 1815, 366, 12312, 596, 1024, 3113, 29973, 319, 1799, 9047, 13566, 29901]

I see that special token `</s>` didn't correctly converted to token id (should be token id = 2) but converted to [29879, 29958]. It is bug?

I saw similar issue in discussion llama.cpp github repo - https://github.com/ggerganov/llama.cpp/issues/1812

antoine-lizee:
@abetlen I second the above. 

I'm not sure where to share this given that it impacts a few different things. In particular, **proper handling of special characters, especially `</s>` is key for any conversation application**.

`llama.cpp` solved the problem only recently (in https://github.com/ggerganov/llama.cpp/pull/3538), and it now works.

I think porting the change to llama-cpp-python will solve a few other problems that are shared around, eg: Chat Formats don't seem to work for instance ( #711 ).

[...]

PR #711, opened 2023-09-14
Configurable Chat Formats

abetlen:
Adds the ability to specify different chat formats. Closes #492 

- [x] Support messages -> prompt formatting
- [x] Handle system messages
- [x] Handle custom stop sequences
- [ ] ~~Handle grammars for function calling formats (functionary)~~
- [x] Support common models (Llama2, Vicuna, Alpaca, etc)

# API

```python
import llama_cpp

# using pre-defined chat model names
model = Llama("vicuna.gguf", chat_format="vicuna")

# using a function that implements the LlamaChatFormat interface
model = Llama("vicuna.gguf", chat_format=llama_cpp.LlamaChatFormatVicuna())
```

# References

- [oobabooga instruction templates](https://github.com/oobabooga/text-generation-webui/tree/main/instruction-templates)
- [FastChat conversations](https://github.com/lm-sys/FastChat/blob/main/fastchat/conversation.py)
- [Functionary schema](https://github.com/MeetKai/functionary/blob/main/functionary/schema.py)

PR #850, opened 2023-10-30
fix: tokenization of special characters:

antoine-lizee:
It should behave like llama.cpp, where most out of the box usages treat special characters accordingly. See https://github.com/abetlen/llama-cpp-python/issues/838#issuecomment-1784966013 for more details.

I checked that with this fix, the vanilla call to `llm.create_completion(temperature=0)` leads to exactly the same results for a simple chat prompt than when using `./main --temp 0` from `llama.cpp` - which it didn't before.

I changed the behaviour also for the embeddings and the LlamaTokenizer. I'm missing context so might be wrong on those, but I figured it would be good to be consistent.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -856,7 +856,7 @@ class Llama:
         data: List[EmbeddingData] = []
         total_tokens = 0
         for index, input in enumerate(inputs):
-            tokens = self.tokenize(input.encode("utf-8"))
+            tokens = self.tokenize(input.encode("utf-8"), special=True)
             self.reset()
             self.eval(tokens)
             n_tokens = len(tokens)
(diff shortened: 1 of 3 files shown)


antoine-lizee:
This should also make Chat Templates work properly ( #711 ) provided that we update a few of them with the eos in the right place (eg: `</s>` for llama2).

antoine-lizee:
@abetlen In case you missed this.

fourdim:
What about removing the empty test.py file?

abetlen:
@antoine-lizee looks good, I'm slightly hesistant to change the default behaviour of the `completion` function, would it be sufficient to only do this for `chat_completion`?

fourdim:
Nope, that will not be sufficient. In my case, I'm infilling codes using `bigcode/starcoder`.
It has special tokens `<fim_prefix>`,  `<fim_suffix>`, `<fim_middle>` to guide starcoder infill the code in the middle rather than the normal completion.
If we set special to False, the model only outputs something random.

abetlen:
I'll go ahead and merge this in as is for now, should have time in the next week to address any issues if this causes breaking changes.

@antoine-lizee thank you for the contribution!

PR #854, opened 2023-10-30
fixed Llama._create_completion suffix check

mtasic85:
We discovered that it was possible to exploit  `Llama._create_completion` suffix parameter to send string-like object which could actually lead to Arbitrary Code Execution.

Here is PoC: https://github.com/tangledgroup/llama-cpp-python-exploit/blob/f99baf1e66b9fa79987c594179ac93eadfb2ca97/exploit.py#L45


abetlen:
@mtasic85 thank you, very good catch!

Issue #788, opened 2023-10-03
Grammar support in llama-cpp-python web server


rounak610:
Does the llama-cpp-python web server support grammar?

abetlen:
@rounak610 not yet but happy to add this!

fvillena:
Please implement grammars!

abetlen:
Implemented in #855 and should be included in v0.2.12

PR #778, opened 2023-10-01
Fix #777, #464

Freed-Wu:
https://github.com/ggerganov/llama.cpp/pull/3424
Move some code about disabling some flags in Apple x86 to the upstream
Use ExternalProjectAdd() to simplify CMakeLists.txt
Change base_path from llama_cpp to llama_cpp/data like other projects:

- https://github.com/ssciwr/clang-format-wheel/blob/99df0fe0145e0af14c38772f1af78b215456e3a3/CMakeLists.txt#L44
- https://github.com/Freed-Wu/astyle-wheel/blob/340cef6dfd5cfa9158a321fe01b36df6a1e826e8/CMakeLists.txt#L12
- https://github.com/scikit-build/cmake-python-distributions/blob/13473f41061f73e599987af8dee5ba262e065769/setup.py#L46
- https://github.com/scikit-build/ninja-python-distributions/blob/e5510f4f725323b9be845a023bdc164718df1ddd/CMakeLists.txt#L149

PR #855, opened 2023-10-31
Grammar support for Web Server

dthuerck:
Created this small PR which just adds a `grammar` parameter to the `completions` endpoint, therefore closing #778 . Of course this violates OpenAI's API definition, but the parameter is optional. Here's an example for it to test (use the json.gnbf from `llama.cpp`):

```
import json
import requests

if __name__ == "__main__":

  prompt = "Please provide me with a valid JSON that encodes the following " \
    "Address: Testine Testerson, 420 Imaginary St, 12345 Star City, USA. " \
    "Make sure yu respect data types and only use one per field. "

  print("######")
  print("# WITHOUT GRAMMAR")
  print("######")

  # simple post request asking for a code problem
  res = requests.post(
    url="http://localhost:8000/v1/completions",
    json={
      "prompt" : f"<s>[INST]{prompt}[/INST]",
      "stop" : ["</s>"],
      "max_tokens" : 256,
      "stream" : True,
      "temperature" : 0.9
    },
    stream=True
  )

  for line in res.iter_lines(decode_unicode=True):
    if line:
      try:
        line = line[6:]
        data = json.loads(line)
        print(data["choices"][0]["text"], end="", flush=True)
      except Exception as e:
        pass

  print("")

  print("######")
  print("# WITH GRAMMAR")
  print("######")

  # now try with using the grammar
  with open("json.gnbf") as f:
    grammar = f.read()

  res = requests.post(
    url="http://localhost:8000/v1/completions",
    json={
      "prompt" : f"<s>[INST]{prompt}[/INST]",
      "stop" : ["</s>"],
      "max_tokens" : 256,
      "stream" : True,
      "temperature" : 0.8,
      "grammar" : grammar
    },
    stream=True
  )

  for line in res.iter_lines(decode_unicode=True):
    if line:
      try:
        line = line[6:]
        data = json.loads(line)
        print(data["choices"][0]["text"], end="", flush=True)
      except Exception as e:
        pass
```

### DIFF
--- a/llama_cpp/server/app.py
+++ b/llama_cpp/server/app.py
@@ -518,6 +518,10 @@ mirostat_eta_field = Field(
     default=0.1, ge=0.001, le=1.0, description="Mirostat learning rate"
 )
 
+grammar = Field(
+    default=None,
+    description="A CBNF grammar (as string) to be used for formatting the model's output."
+)
 
 class CreateCompletionRequest(BaseModel):
     prompt: Union[str, List[str]] = Field(
@@ -533,6 +537,7 @@ class CreateCompletionRequest(BaseModel):
     mirostat_mode: int = mirostat_mode_field
     mirostat_tau: float = mirostat_tau_field
     mirostat_eta: float = mirostat_eta_field
+    grammar: Optional[str] = None
     echo: bool = Field(
         default=False,
         description="Whether to echo the prompt in the generated text. Useful for chatbots.",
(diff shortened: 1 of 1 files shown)


Freed-Wu:
@dthuerck you should close #788 not #778

aabbi:
Thanks for adding this. The parameter can be passed as an `extra_body` parameter when using the openai client. Would it make sense to add this in the documentation somewhere (I had to dig a little bit to figure out how to do this) ?
Example usage
```python
llm_uri = "http://localhost:8000/v1"
llm_client = OpenAI(base_url=llm_uri, api_key="xx")
modelname = "llama-2-7b-chat.Q6_K.gguf"
with open("json.gnbf") as f:
    grammar = f.read()
resp = llm_client.chat.completions.create(messages=prompt, model=modelname, extra_body={"grammar": grammar})

PR #869, opened 2023-11-03
Update llama_chat_format.py

earonesty:
properly formal llama2 with first-message prompt embedded

probably, this is not the right PR 

can fold it into the general templating system instead


abetlen:
Awesome, thank you!

PR #913, opened 2023-11-15
fix: get system message from messages for all prompt formats

julianullrich99:
Some prompt formats were missing the _get_system_message() call


julianullrich99:
bumb @abetlen

PR #920, opened 2023-11-16
fix: openblas_simple missing pkg-config to build

ZisisTsatsas:
Adding pkg-config to Dockerfile for openblas_simple


abetlen:
Thank you

Issue #911, opened 2023-11-15
Add min p arg to server


ArtyomZemlyak:
Add `min p` arg to server
Related to :
https://github.com/ggerganov/llama.cpp/pull/3841

iandennismiller:
It looks like `min p` was added to llama-cpp-python recently:

https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama_cpp.py#L1560

https://github.com/abetlen/llama-cpp-python/commit/d808fd436c2f0f4df1f491e99c8bcb3b98b4bcbe?diff=split#diff-9184e090a770a03ec97535fbef520d03252b635dafbed7fa99e59a5cca569fbcR1568

LorenzoBoccaccia:
I think it's missing from llama.py sampling

ddh0:
Second this! It looks very promising and I'm excited.

ddh0:
Pinging @abetlen, I finished this. See here: https://github.com/ddh0/llama-cpp-python/commit/e8e05bbb0f4100d33b65f9595c81e341814c8097

Apologies if this isn't the right way to submit a fix, I'm new to github. But I've tested it and it's working as expected. Let me know if I can help in any way.

tk-master:
@ddh0 bad timing, I just made this pr :P https://github.com/abetlen/llama-cpp-python/pull/921

m-from-space:
> tested at temperature 1.6 and 3.0 the assistant remains creative and repetition issues are gone even with a lower repetition penalty. at 3.0 seems to be too creative and basically ignore facts passed in the context, but remains coherent and inteligible.
> 
> still I think it should be disabled by default for backcompat

Hey there, I just saw that this is now implemented. How exactly do I activate `min_p` sampling when calling the model? Is it activated by default?

tk-master:
> > tested at temperature 1.6 and 3.0 the assistant remains creative and repetition issues are gone even with a lower repetition penalty. at 3.0 seems to be too creative and basically ignore facts passed in the context, but remains coherent and inteligible.
> > still I think it should be disabled by default for backcompat
> 
> Hey there, I just saw that this is now implemented. How exactly do I activate `min_p` sampling when calling the model? Is it activated by default?

the same way you'd change the temperature, top_k etc.. it is activated with a small value by default

m-from-space:
> the same way you'd change the temperature, top_k etc.. it is activated with a small value by default

[...]

PR #921, opened 2023-11-16
Added support for min_p

tk-master:
My small contribution to this great project.

Ref: https://github.com/ggerganov/llama.cpp/pull/3841

Closes: https://github.com/abetlen/llama-cpp-python/issues/911


abetlen:
@tk-master thank you for the contribution!

Issue #912, opened 2023-11-15
0.18 (and others?) no longer use GPU properly.


pseudotensor:
https://github.com/jllllll/llama-cpp-python-cuBLAS-wheels/issues/21#issuecomment-1809000662

If I only install GPU version, fails needing llama cpp python.  With both installed, never uses GPU, no mention of offlload as usual, etc.

JohanAR:
I've got a suspicion that it might have been caused by multiple "import llama_cpp" having been added recently, as opposed to using relative imports within the package. AFAIK jlllllll renames the package to llama_cpp_cuda to allow ooba's webui to include both.

pseudotensor:
Sounds plausible.

cognitivetech:
> I've got a suspicion that it might have been caused by multiple "import llama_cpp" having been added recently, as opposed to using relative imports within the package. AFAIK jlllllll renames the package to llama_cpp_cuda to allow ooba's webui to include both.

except for the fact that when I try alone in python shell it still doesn't work?

JohanAR:
@cognitivetech the issue title is a bit too broad and the description a bit short, but we're talking about using Ooba's webui with jllllll's pre-build wheel files which have renamed the package to llama_cpp_cuda. It's perhaps somewhat more clear if you follow the link to the other repo and read the issue and comments there.

If you're having problems with getting llama-cpp-python to work with ROCm I think it's most likely unrelated, and I'd recommend opening a separate issue for that.

PR #922, opened 2023-11-16
Replace all absolute imports of llama_cpp in llama_cpp

JohanAR:
Hopefully solves #912 and allows this package to be renamed without dependence on llama_cpp


JohanAR:
And in case anyone is wondering _why_ someone would want to rename the package, https://github.com/jllllll/llama-cpp-python-cuBLAS-wheels provides a CUDA enabled version renamed to llama_cpp_cuda to allow https://github.com/oobabooga/text-generation-webui to load both the CUDA and CPU versions simultanously.

Issue #906, opened 2023-11-13
Serve multiple models with [server]


bioshazard:
lcp[server] has been excellent. And I can host two models by running a second instance.

I'd like to be able to serve multiple models with a single instance of the OpenAI-compatible server and switch between them based on alias-able `model` in the query payload. My use case is to serve a code model and bakllava at the same time.

I am going to see about attempting to PR an Nginx configuration example that would reverse proxy to two instances based on the `model` in the POST body, but first order support would be great. If no one picks this up, I might attempt a PR of first order support early '24.

bioshazard:
Might also help to serve another model specifically for embeddings than to rely on whatever my latest 7B obsession does.

NewtonTrendy:
I would like this too!

abetlen:
Implemented in #931 next up is #736

bioshazard:
Thank you!! And wow #736 is a huge idea.

Issue #736, opened 2023-09-18
[FR] Auto-load/unload models to save memory when unused


NightMachinery:
**Is your feature request related to a problem? Please describe.**
In a shared machine, or even a personal one, it's better to automatically free up the memory used when the model is not in use.

**Describe the solution you'd like**
Some `idle_before_unload=600` parameter that unloads the model after that many seconds of being idle. The model should be reloaded automatically whenever it is used again.

**Additional context**
- https://github.com/abetlen/llama-cpp-python/issues/302

PR #931, opened 2023-11-21
[Feat] Multi model support

D4ve-R:
# Added multi model support

## Changes
* Llama class handles chat_handler initialization based on chat_format internal in __init__()
* Llama class handles cache initialization internal in __init__()
* Change Request parameter model in api routes from optional to mandatory
* Add MultiLlama class that handles multiple models
* Settings in own file

## Description
MultiLlama class handles multiple models in the given dir.
It loads models lazy based the model request param, e.g. will take some time at first api call or after model change.

## Usage
```
/models
|__mistral-orca.gguf
|__mistral-instruct-v0.1.gguf
```
```
$ python3 -m llama_cpp.server --model /models

# works too, but model request param still mandatory
$ python3 -m llama_cpp.server --model /models/mistral-orca.gguf
```
```python
from openai import OpenAI
client = OpenAI(api_key="sk-oss4win")
chat_completion = client.chat.completions.create(
    model="mistral-orca", # or mistral-instruct-v0.1,
    messages=[...],
)
```


D4ve-R:
Closes #906

limoncc:
this is that i want

abetlen:
Hi @D4ve-R thanks for the PR. Just a couple of notes:

- Avoid the `llama.py` changes that add the clip path parameter, I think the chat handler approach is best right now while the api stabilises.
- Don't change `model` to mandatory, this will cause the server to break for many users, it's better to have a default model set by the user or default to the first model.
- For loading multiple models I think it would make sense to use a json file with parameters and 1 settings object per model

Something like:
```json
{
   "host": "...",
   "port": "...",
   "models: [ {
       "model": "models/mistral-7b/...",
       "chat_format": "mistral"
   }]
}
```

This way we can re-use the pydantic settings object inside some kind of config file settings object that just parses the json.

D4ve-R:
Hi @abetlen, thanks for the feedback & awesome project! 
> * Avoid the `llama.py` changes that add the clip path parameter, I think the chat handler approach is best right now while the api stabilises.
> * Don't change `model` to mandatory, this will cause the server to break for many users, it's better to have a default model set by the user or default to the first model.

I will revert the changes concerning the clip_model & model param.

> * For loading multiple models I think it would make sense to use a json file with parameters and 1 settings object per model
> 
> Something like:
> 
> ```json
> {
>    "host": "...",
>    "port": "...",
>    "models": [{
>        "model": "models/mistral-7b/...",
>        "chat_format": "mistral"
>    }]
> }
> ```
> 
> This way we can re-use the pydantic settings object inside some kind of config file settings object that just parse

I was thinking about something like this too, i will try to make something work

D4ve-R:
Alternatively/additionally the app could expose a new endpoint "/settings", to set settings via api, @abetlen what are your thoughts on that?

D4ve-R:
Okay, so loading individual settings as json works now.
If loading [model].json fails, it will fallback to using the global settings.
### Usage:
```
/models
|__mistral-orca.gguf
|__mistral-orca.json
|__mistral-instruct-v0.1.gguf
```

abetlen:
Great work.

> Alternatively/additionally the app could expose a new endpoint "/settings", to set settings via api, @abetlen what are your thoughts on that?

Admin routes should be easy to set up given that all the settings are just pydantic models but probably out of scope for this PR.

> Okay, so loading individual settings as json works now. If loading [model].json fails, it will fallback to using the global settings.
> 
> ### Usage:
> ```
> /models
> |__mistral-orca.gguf
> |__mistral-orca.json
> |__mistral-instruct-v0.1.gguf
> ```
> 
> [model].json
> 
> ```json
> {
>   "n_gpu_layers": 0,
>   "main_gpu": 0,
>   "tensor_split": null,
>   "vocab_only": false,
>   "use_mmap": true,
>   "use_mlock": true,
>   "seed": 4294967295,
>   "n_ctx": 2048,
>   "n_batch": 512,
>   "n_threads": 4,
>   "n_threads_batch": 4,
>   "rope_scaling_type": -1,
>   "rope_freq_base": 0.0,
>   "rope_freq_scale": 0.0,
>   "yarn_ext_factor": -1.0,
>   "yarn_attn_factor": 1.0,
>   "yarn_beta_fast": 32.0,
>   "yarn_beta_slow": 1.0,
>   "yarn_orig_ctx": 0,
>   "mul_mat_q": true,
>   "f16_kv": true,
>   "logits_all": true,
>   "embedding": true,
>   "last_n_tokens_size": 64,
>   "lora_base": null,
>   "lora_path": null,
>   "numa": false,
>   "chat_format": "llama-2",
>   "clip_model_path": null,
>   "cache": false,
>   "cache_type": "ram",
>   "cache_size": 2147483648,
>   "verbose": true
> }
> ```

I think we should stick to a single config file that can act as a replacement for the server's cli arguments. ie you either specify settings for a single model on the cli or you pass a `--config <path>` which points to a single `config.json`

So something more like

```
> /models
> |__mistral-7b.gguf
> |__open-hermes-2.5-7b.gguf
```

`config.json`
```json
{
    "models": [{
         "model": "models/mistral-7b.gguf",
         "model_alias": "text-davinci-003"
   }, {
         "model": "models/open-hermes-2.5-7b.gguf",
         "model_alias": "gpt-3.5-turbo"
   }
}
```

The advantage is that you only need to manage / change a single config file so you can swap configs easily and we can also start to add server global options at the root of this document.

D4ve-R:
@abetlen, you are right that makes way more sense.
It works now!
```sh
# use as before
python -m llama_cpp.server --model /models/mymodel.gguf [--args]
# use with multiple models all with default config or from arguments
python -m llama_cpp.server --model /models [--args]
# use multiple models with config file
python -m llama_cpp.server  --config /config.json
```
In the config file, default params and individual params can be set.
Missing values in the individual config will get replaced by specified default config values.
Also supports multi aliasing with different params.
So the config file can look like this:
```json
{
    "model": "/models",
    "n_gpu_layers": -1,
    "verbose": true,
    "models": [
        {
            "model": "/models/mistral-7b-instruct-v0.1.Q4_0.gguf",
            "model_alias": "mistral", 
            "verbose": false
        },
       {
            "model": "/models/mistral-7b-instruct-v0.1.Q4_0.gguf",
            "model_alias": "mistral-long-ctx",
            "n_ctx": 4096
        },
        {
            "model": "/models/bakllava-v1.gguf",
            "model_alias": "bakllava",
            "chat_format": "llava-1-5",
            "clip_model_path": "/models/mmproject.gguf"
        }
    ]
  }

```

abetlen:
@D4ve-R this is awesome, great contribution, will review it tonight.

I think after this the next big thing is #736 , would be very powerful to define a config with a few different models and have them load / unload as you use the API.

llama_cpp/llama.py
```diff
@@ -759,6 +759,10 @@ def __init__(
         # Chat Format Params
         chat_format: str = "llama-2",
         chat_handler: Optional[llama_chat_format.LlamaChatCompletionHandler] = None,
+        # Cache
+        cache: bool = False,
+        cache_type: str = "ram",
+        cache_size: int = 2 << 30,
```
abetlen, line 765: Can we keep these cache changes out of this PR, I'm currently working on the cache as part of batch processing and want to avoid any conflicts.

llama_cpp/llama.py
```diff
@@ -958,6 +965,17 @@ def __init__(
             (n_ctx, self._n_vocab), dtype=np.single
         )
 
+        if cache:
```
abetlen, line 968: same as above

llama_cpp/server/app.py
```diff
@@ -882,19 +653,16 @@ class ModelList(TypedDict):
...
 ) -> ModelList:
-    assert llama is not None
     return {
         "object": "list",
         "data": [
             {
-                "id": settings.model_alias
-                if settings.model_alias is not None
-                else llama.model_path,
+                "id": model,
```
abetlen, line 662: This should probably have similar logic to the left side, id should only dump the "name" of the model which can be used for selecting the model in completion and chat_completion requests.

llama_cpp/server/app.py
```diff
@@ -820,16 +591,16 @@ class CreateChatCompletionRequest(BaseModel):
 async def create_chat_completion(
     request: Request,
     body: CreateChatCompletionRequest,
-    llama: llama_cpp.Llama = Depends(get_llama),
-    settings: Settings = Depends(get_settings),
+    llama: Llama = Depends(get_llama),
```
abetlen, line 594: Should fix the type signature here right? Since this depends now yields an `Optional[LlamaProxy]`

llama_cpp/server/__main__.py
```diff
@@ -22,12 +22,16 @@
...
 import argparse
 from typing import List, Literal, Union
 
 import uvicorn
 
-from llama_cpp.server.app import create_app, Settings
+from llama_cpp.server.app import create_app
+from llama_cpp.server.settings import Settings, ServerSettings, set_settings
+
+EXE_NAME = 'llama_server'
```
abetlen, line 34: For now I think this should probably be `llama_cpp.server`

abetlen (changes requested):

llama_cpp/server/app.py
```diff
@@ -882,19 +653,16 @@ class ModelList(TypedDict):
...
             {
-                "id": settings.model_alias
-                if settings.model_alias is not None
-                else llama.model_path,
+                "id": model_alias,
                 "object": "model",
                 "owned_by": "me",
                 "permissions": [],
-            }
+            } for model_alias in llama
```
D4ve-R, line 666: This always returns the model alias, if no model_alias was configured, it defaults to the filename without extension

abetlen: I figured that was probably just my confiusion

pyproject.toml
```diff
@@ -63,6 +63,9 @@ sdist.include = [".git", "vendor/llama.cpp/.git"]
 provider = "scikit_build_core.metadata.regex"
 input = "llama_cpp/__init__.py"
 
+[project.scripts]
```
abetlen, line 66: Can we actually keep this out for now, the `pyhton3 -m llama_cpp.server` approach is okay for the time being and we may want to think through the name a little more for adding an executable.

abetlen:
Okay I've got all the merge conflicts resolved and I did some general refactoring of the server submodule. It's a lot lighter than the original PR but that was necessary to avoid any breaking changes and I plan to reintegrate the additional features one at a time.

# Overview

You can now pass a `--config_file` or use the `CONFIG_FILE` environment variable to load a server config file that can include multiple models. Here's an example of the config file

```json
{
    "host": "0.0.0.0",
    "port": 8080,
    "models": [
        {
            "model": "models/OpenHermes-2.5-Mistral-7B-GGUF/openhermes-2.5-mistral-7b.Q4_K_M.gguf",
            "model_alias": "gpt-3.5-turbo",
            "chat_format": "chatml",
            "n_gpu_layers": -1,
            "offload_kqv": true,
            "n_threads": 12,
            "n_batch": 512,
            "n_ctx": 2048
        },
        {
            "model": "models/OpenHermes-2.5-Mistral-7B-GGUF/openhermes-2.5-mistral-7b.Q4_K_M.gguf",
            "model_alias": "gpt-4",
            "chat_format": "chatml",
            "n_gpu_layers": -1,
            "offload_kqv": true,
            "n_threads": 12,
            "n_batch": 512,
            "n_ctx": 2048
        },
        {
            "model": "models/ggml_llava-v1.5-7b/ggml-model-q4_k.gguf",
            "model_alias": "gpt-4-vision-preview",
            "chat_format": "llava-1-5",
            "clip_model_path": "models/ggml_llava-v1.5-7b/mmproj-model-f16.gguf",
            "n_gpu_layers": -1,
            "offload_kqv": true,
            "n_threads": 12,
            "n_batch": 512,
            "n_ctx": 2048
        },
        {
            "model": "models/mistral-7b-v0.1-GGUF/ggml-model-Q4_K.gguf",
            "model_alias": "text-davinci-003",
            "n_gpu_layers": -1,
            "offload_kqv": true,
            "n_threads": 12,
            "n_batch": 512,
            "n_ctx": 2048
        },
        {
            "model": "models/replit-code-v1_5-3b-GGUF/replit-code-v1_5-3b.Q4_0.gguf",
            "model_alias": "copilot-codex",
            "n_gpu_layers": -1,
            "offload_kqv": true,
            "n_threads": 12,
            "n_batch": 1024,
            "n_ctx": 9216
        }
    ]
}
```

The server selects the appropriate model based on the `model` key passed to it and the `model_alias` defined in the config fiile.
If the selected model is not loaded already it's loaded on-the-fly.
The server currently only handles one single loaded model at a time and will automatically.

# Next steps
- [ ] Support for setting default model fields to make the config file less verbose.
- [ ] Support for additional config file formats (yaml / toml / ini) so the config can support comments and multiline strings.
- [ ] Support a `models_dir` or auto-generating a config file from a script to make it easier to get up and running with the config files.
- [ ] Automatically unload a model after some period of time.
- [ ] Documentation and examples.

abetlen:
Here's a handy bash one-liner to traverse all of the `.gguf` files in a directory and generate a skeleton config.

```bash
find path/to/directory -name '*.gguf' -size +1k -printf '%p\n' | jq -R -s -c '{"models": [split("\n")[:-1][] | {"model": .}]}'
```

PR #947, opened 2023-11-25
fix: Typo in the Open Orca chat format #874

gardner:
The string assignment for `system_message` was broken into two parts without proper concatenation. 

This change concatenates these two string parts properly.


abetlen:
Thank you!

PR #952, opened 2023-11-28
Fix #891

### LINKED ISSUE #891: llama_cpp conflict with transformers library
Hi everyone, 

I'm currently trying to run the example script provided in the readme, but I noticed a Type error exception is thrown (and ignored) when importing the `transformers==4.35.0` library in the same file. 

# Expected Behavior

The code snippet in the next section should generate an output without throwing any error. 

# Current Behavior

By running the following code snipped, I get an exception after the output is correctly generated. I'm aware the error is ignored but I would like to understand whether that's the expected behaviour or if there is something else I'm missing. Moreover, notice that removing  `import transformers` produces the correct output. 

```python
import transformers
from llama_cpp import Llama

if __name__ == "__main__":

    llm = Llama(model_path="./models/7B/llama-model.gguf")
    output = llm("Q: Name the planets in the solar system? A: ",
                 max_tokens=32, stop=["Q:", "\n"], echo=True)

    print(output)

```
And I get the following error

```
Exception ignored in: <function _LlamaContext.__del__ at 0x7f4cfd08a3b0>
  File "/home/project/my_env/lib/python3.10/site-packages/llama_cpp/llama.py", line 418, in __del__
TypeError: 'NoneType' object is not callable
Exception ignored in: <function _LlamaModel.__del__ at 0x7f4cfd089480>
  File "/home/project/my_env/lib/python3.10/site-packages/llama_cpp/llama.py", line 238, in __del__
TypeError: 'NoneType' object is not callable
Exception ignored in: <function _LlamaBatch.__del__ at 0x7f4cfd08b7f0>
  File "/home/project/my_env/lib/python3.10/site-packages/llama_cpp/llama.py", line 664, in __del__
TypeError: 'NoneType' object is not callable
```

# Environment and Context

- x86_64 GNU/Linux
- Python 3.10.9
- GNU Make 4.2.1
- g++ 9.4.0

kddubey:
Fix https://github.com/abetlen/llama-cpp-python/issues/891

As noted in the [Python docs](https://docs.python.org/3/reference/datamodel.html#object.__del__), objects can get nullified on interpreter shutdown in `__del__`. The solution follows the same pattern [here](https://github.com/abetlen/llama-cpp-python/blob/ad963a0961b04195d6e360ac041b4c558495e0da/llama_cpp/_utils.py#L6-L7)

## How has this been tested?

1. Install a later version of transformers with torch:

   ```
   python -m pip install "transformers[torch]>=4.35.0"
   ```

2. Checkout main

   ```
   git checkout main
   ```

3. Verify that this command prints out the ignored Exception:

   ```
   python -c "import transformers; from llama_cpp import Llama; llm = Llama(model_path='./some.gguf', verbose=False)"
   ```

4. Checkout this branch

   ```
   git checkout ignore-warnings-on-shutdown
   ```

5. Verify that the same command **does not** print out the ignored Exception:

   ```
   python -c "import transformers; from llama_cpp import Llama; llm = Llama(model_path='./some.gguf', verbose=False)"
   ```

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -214,6 +214,8 @@ class _LlamaModel:
     NOTE: For stability it's recommended you use the Llama class instead."""
 
     _llama_free_model = None
+    # NOTE: this must be "saved" here to avoid exceptions when calling __del__
+    suppress_stdout_stderr = suppress_stdout_stderr
 
     def __init__(
         self,
@@ -237,7 +239,7 @@ class _LlamaModel:
             )
 
     def __del__(self):
-        with suppress_stdout_stderr(disable=self.verbose):
+        with self.suppress_stdout_stderr(disable=self.verbose):
             if self.model is not None and self._llama_free_model is not None:
                 self._llama_free_model(self.model)
                 self.model = None
@@ -399,6 +401,8 @@ class _LlamaContext:
     NOTE: For stability it's recommended you use the Llama class instead."""
 
     _llama_free = None
+    # NOTE: this must be "saved" here to avoid exceptions when calling __del__
+    suppress_stdout_stderr = suppress_stdout_stderr
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


kddubey:
@abetlen if this looks good, apply the same changes to your (awesome!) [PR here](https://github.com/abetlen/llama-cpp-python/pull/951/files#diff-a5923388174e5c4c6b7aa2cd2c4645f5666833e66773dcad9a71f8a0e5768dd3)

abetlen:
This is perfect, thank you!

PR #955, opened 2023-11-29
Update shared library import & license compliance

D4ve-R:
Install shared libs from llama.cpp into `/lib` and copy the license to ensure compliance.


abetlen:
@D4ve-R thank you!

I think we also need to update the `clean` target (any maybe some others) in the Makefile but otherwise looks good to me.

PR #957, opened 2023-11-30
Generic Function Calling

abetlen:
[Reference](https://github.com/abacaj/openhermes-function-calling)

Tested this method in a notebook and works quite well.

TODO:
- [x] Adjust the prompt to support mulit-turn conversations
- [x] Use gramamr constrained sampling
- [x] Add custom `ChatCompletionHandler`
- [x] Support legacy function calling API
- [x] Support multi-tool auto calls (hard)
- [ ] Support for streaming muti-tool auto calls
- [x] Test on Function Calling notebook and instructor cookbook


Maximilian-Winter:
A little framework for structured output, chat and function calling: 
https://github.com/Maximilian-Winter/llama-cpp-agent

Maximilian-Winter:
@abetlen Could you tell me if this is useful for you? I can also integrate just the grammar generator into llama-cpp-python and make a pull request.

teleprint-me:
I am very interested in this. Will be reviewing when I have some time.

vriesdemichael:
Be careful how you promote this chat handler. With this prompt it is only suitable for function calling, which might be confusing if you expect the handler to provide an OpenAI compatible experience with tool calls only when needed.

I am also not entirely sure if openhermes mistral was trained on these specific instructions or that it would work with other models as well.

If it just good prompt engineering to coerce a tool call, then langchain already has some tools to do so using reAct
https://github.com/williamcotton/empirical-philosophy/blob/main/articles/how-react-prompting-works.md

abetlen:
Just missing automatic multi-tool use now.

Issue #872, opened 2023-11-05
suppress_stdout_stderr failing in notebooks


slundberg:
# Expected Behavior

Loading a model with verbose=False should not print things.

# Current Behavior

verbose=False has no effect in notebooks (I tested with VS Code notebooks, but I think it is ipykernel related)

# Steps to Reproduce

Just load a model with verbose=False and observe output in a notebook environment.

# How to fix

The patch given in the ipykernel issue below fixes the issue for me (but requires me to do my own override of stdout and stderr since the patch also causes `suppress_stdout_stderr` to crash as currently written). A real fix might involve integrating this into `suppress_stdout_stderr` directly.

https://github.com/ipython/ipykernel/issues/795

thanks!

slundberg:
I got a nicer version of the patch working that I will share here is case it helps others:

```python
class normalize_ipykernel_stdout_stderr():
    '''Remaps stdout and stderr back to their normal selves from what ipykernel did to them.
    
    Based on: https://github.com/ipython/ipykernel/issues/795
    '''

    def __enter__(self):
        normal_stdout = sys.__stdout__.fileno()
        self.restore_stdout = None
        if getattr(sys.stdout, "_original_stdstream_copy", normal_stdout) != normal_stdout:
            self.restore_stdout = sys.stdout._original_stdstream_copy
            sys.stdout._original_stdstream_copy = normal_stdout

        normal_stderr = sys.__stderr__.fileno()
        self.restore_stderr = None
        if getattr(sys.stderr, "_original_stdstream_copy", normal_stderr) != normal_stderr:
            self.restore_stderr = sys.stderr._original_stdstream_copy
            sys.stderr._original_stdstream_copy = normal_stderr

    def __exit__(self, exc_type, exc_value, traceback):
        if self.restore_stdout is not None:
            sys.stderr._original_stdstream_copy = self.restore_stdout
        if self.restore_stderr is not None:
            sys.stderr._original_stdstream_copy = self.restore_stderr
```

given the above context manager you can now wrap a model load and things work like you would expect:

```python
with normalize_ipykernel_stdout_stderr():
    llama_cpp.Llama(model_path=model, verbose=False)
```

abetlen:
@slundberg yes been meaning to dig into this issue, thank you so much for the fix!

abetlen:
I haven't been able to get a version that consistently works in all environments, if anyone can give this a shot it would be much appreciated.

brandonrobertz:
[...]

PR #961, opened 2023-11-30
Bug fixing from llava and chathandler about UnsupportedOperation: fileno

zocainViken:
def __enter__(self):
        if self.disable:
            return self

        # Check if sys.stdout and sys.stderr have fileno method
        if not hasattr(self.sys.stdout, 'fileno') or not hasattr(self.sys.stderr, 'fileno'):
            return self  # Return the instance without making changes

    def __exit__(self, *_):
        if self.disable:
            return

        # Check if sys.stdout and sys.stderr have fileno method
        if hasattr(self.sys.stdout, 'fileno') and hasattr(self.sys.stderr, 'fileno'):

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -47,4 +47,4 @@ jobs:
       - name: Publish to GitHub Tag
         if: steps.docker_build.outputs.digest && startsWith(github.ref, 'refs/tags/')
         run: |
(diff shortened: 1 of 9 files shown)


zocainViken:
[suppress_stdout_stderr #872](https://github.com/abetlen/llama-cpp-python/issues/872)

abetlen:
@zocainViken thank you! Does this now resolve #872 ?

PR #967, opened 2023-12-02
README.md multimodal params fix

zocainViken:
multi modal params fix: add logits = True -> to make llava work


abetlen:
@zocainViken thank you!

PR #991, opened 2023-12-09
Replace logits_to_logprobs implementation with numpy equivalent to llama.cpp

player1537:
See #990. This change makes the logits_to_logprobs function equivalent to the version in the llama.cpp repository. It uses numpy so it's much faster than the previous version.

*Edit:* I should add that I'm not 100% confident in this change. I'd appreciate if someone else could test this themselves too.


abetlen:
@player1537 perfect, thank you very much!

PR #994, opened 2023-12-11
Add support for running the server with SSL

rgerganov:


abetlen:
@rgerganov thank you very much!

PR #1002, opened 2023-12-12
Fix logits_to_logprobs for 2-D and 3-D logits

kddubey:
The [implementation in main](https://github.com/abetlen/llama-cpp-python/blob/8e44a32075de4aba2fc9877d4a2a34a0e7314c0d/llama_cpp/llama.py#L2282-L2290) (from [this PR](https://github.com/abetlen/llama-cpp-python/pull/991)) only works for 1-D logits. It silently fails for 2-D or 3-D logits. The implementation in this PR works out-of-the-box for 1-D, 2-D, and 3-D logits. (3-D is possible in the future w/ batch inference and `logits_all=True`.) This feature might be useful b/c there are [some](https://github.com/abetlen/llama-cpp-python/blob/8e44a32075de4aba2fc9877d4a2a34a0e7314c0d/llama_cpp/llama.py#L1785-L1787) places in the code where we can save time by vectorizing / not converting data to lists. I'll do that in a future PR.

The minimal and sufficient fix is to set `axis=-1` in the `np.max` call, and set `keepdims=True` in the `np.sum` call. I decided to instead go with a more robust implementation. It's almost copy-pasted from [`scipy.special.log_softmax`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.log_softmax.html). I decided against adding `scipy` as a required dependency b/c it's not lightweight—the latest version is ~37 MB.

### Script

1. Install the new test dependency, `scipy`, which contains a correct implementation

   ```bash
   python -m pip install scipy
   ```

2. Checkout main

   ```bash
   git checkout main
   ```

3. Run this script in main to verify that the current implementation is silently wrong for 2-D logits

   ```python
   from __future__ import annotations
   
   import numpy as np
   from scipy.special import log_softmax
   
   from llama_cpp import Llama
   
   atol = 1e-3  # intentionally set to be loose when testing the impl in main
   size = (2, 3)
   logits: list = (
       (-np.random.uniform(low=0, high=60, size=size)).astype(np.single).tolist()
   )
   
   logprobs = Llama.logits_to_logprobs(logits)
   logprobs_correct = log_softmax(logits, axis=-1)
   assert np.allclose(logprobs, logprobs_correct, atol=atol)
   
   ```

4. Checkout this branch

   ```bash
   git checkout kddubey/fix-logits-to-logprobs
   ```

5. Run the same script with `atol=1e-6`. No error should be raised.

### New unit tests

```
pytest tests/test_llama.py -k test_logits_to_logprobs
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,11 +33,12 @@ server = [
     "fastapi>=0.100.0",
     "pydantic-settings>=2.0.1",
     "sse-starlette>=1.6.1",
-    "starlette-context>=0.3.6,<0.4"
+    "starlette-context>=0.3.6,<0.4",
 ]
 test = [
     "pytest>=7.4.0",
     "httpx>=0.24.1",
+    "scipy>=1.10",
 ]
 dev = [
     "black>=23.3.0",
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -2289,7 +2289,7 @@ class Llama:
             logits_maxs[~np.isfinite(logits_maxs)] = 0
         elif not np.isfinite(logits_maxs):
             logits_maxs = 0
-        subtract_maxs = logits - logits_maxs
+        subtract_maxs = np.subtract(logits, logits_maxs, dtype=np.single)
         exp = np.exp(subtract_maxs)
         # Suppress warnings about log of zero
         with np.errstate(divide='ignore'):
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -771,7 +771,7 @@ class Llama:
         **kwargs,  # type: ignore
     ):
         """Load a llama.cpp model from `model_path`.
-    
+
         Examples:
             Basic usage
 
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -33,11 +33,12 @@ server = [
     "fastapi>=0.100.0",
     "pydantic-settings>=2.0.1",
     "sse-starlette>=1.6.1",
-    "starlette-context>=0.3.6,<0.4"
+    "starlette-context>=0.3.6,<0.4",
 ]
 test = [
     "pytest>=7.4.0",
     "httpx>=0.24.1",
+    "scipy>=1.10",
```
kddubey, line 41: This is the oldest version compatible with [`numpy>=1.20.0`](https://github.com/abetlen/llama-cpp-python/blob/8e44a32075de4aba2fc9877d4a2a34a0e7314c0d/pyproject.toml#L16)

source: https://docs.scipy.org/doc/scipy/dev/toolchain.html#numpy

abetlen:
@kddubey thank you, yes that's a good idea wrt vectorizing the logits -> logprobs calculation

PR #1011, opened 2023-12-14
Implement GGUF metadata KV overrides

### LINKED ISSUE #1084: Add KV overrides support to `Llama` class and server

phiharri:
Quick implementation of KV overrides which accepts strings in llama.cpp `--kv_overrides KEY=TYPE:VALUE ..` format. For example `llama.expert_used_count=int:3`  Multiple overrides may be space separated.

Not that familiar with ctypes or the preferred types to use here.

Closes #1084


abetlen:
Hey @phiharri thanks for the contribution, I have a few changes to make but the overall api seems correct. I think the kv overrides argument to the `Llama` class should just be a dictionary and the types should be based on the python types but the string based approach for the cli args looks good.

abetlen:
Looks good now, @phiharri thank you for the contribution!

PR #1012, opened 2023-12-14
Avoid "LookupError: unknown encoding: ascii" when open() called in a …

yieldthought:
…destructor

The existing code often causes "LookupError: unknown encoding: ascii" when open() called in a destructor. Saving open in self.open is not enough to avoid this. Instead, we can avoid reopening /dev/null every time by doing it once when the module is loaded.

### DIFF
--- a/llama_cpp/_utils.py
+++ b/llama_cpp/_utils.py
@@ -1,11 +1,15 @@
 import os
 import sys
 
+import sys, traceback
+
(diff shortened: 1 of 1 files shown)


FGRibreau:
Thanks a lot! I had the same issue :)

abetlen:
@yieldthought thank you for the contribution!

PR #1015, opened 2023-12-15
Bug fixed with n_ctx=0

### LINKED ISSUE #988: Llama(n_ctx = 0) doesn't work
# Prerequisites

- [x] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [x] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

As stated in the docs, creating Llama with n_ctx=0 should default to the model's trained context length and work.

# Current Behavior

Instead, llama_cpp crashes after loading the model.

# Environment and Context

Linux 6.2 kernel, Python 3.11, latest llama_cpp instaled with CUBLAS support.

# Failure Information / Steps to Reproduce

```
import llama_cpp
model = llama_cpp.Llama(model_path="../models/llama-2-7b-chat.Q4_K_M.gguf", n_ctx=0)
print(model.n_ctx())
print(model("The quick brown fox jumps ", stop=["."])["choices"][0]["text"])
```

# Extra Info

I think this is related with the following line inside Llama.\__init\__(), effectively setting self.n_batch=0, since n_ctx is 0:
```
        self.n_batch = min(n_ctx, n_batch)  # ???
```

Removing this line and setting self.n_batch = n_batch, avoids the crash. The above code prints the model's trained n_ctx. However later raises an exception during inference and doesn't complete it:

``` 
ValueError: could not broadcast input array from shape (39,) into shape (0,)
```

DanieleMorotti:
If the ```n_ctx``` parameter is set to 0 the function should use the maximum context length of the selected model, but it didn't work. There was a problem with the initialization of this parameter and a related problem with ```n_batch```.

I know that the code is not the best one, but in order to get the model information about the context I needed to add it after the creation of the model instance at line 923 of the file  ```llama.py```.
```python
self._model = _LlamaModel(
    path_model=self.model_path, params=self.model_params, verbose=self.verbose
)
``` 

Unfortunately, different objects were already initialized, therefore in the fix I had to change the ```n_ctx```, ```self.n_batch```, ```self.context_params.n_ctx``` and ```self.context_params.n_batch``` variables even if they already had a value.

Tell me if you find a smarter or more elegant solution to change the code and i will implement it. 

This change should also fix #988.

Thank you

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -923,6 +923,12 @@ class Llama:
         self._model = _LlamaModel(
             path_model=self.model_path, params=self.model_params, verbose=self.verbose
         )
+        # Set the default value for the context and correct the batch
+        if n_ctx == 0:
+            n_ctx = self._model.n_ctx_train()
+            self.n_batch = min(n_ctx, n_batch)
+            self.context_params.n_ctx = self._model.n_ctx_train()
+            self.context_params.n_batch = self.n_batch
 
         self._ctx = _LlamaContext(
             model=self._model,


DanieleMorotti:
Another approach can be to use the code in the [llama.cpp repo](https://github.com/ggerganov/llama.cpp/blob/master/gguf-py/scripts/gguf-dump.py) to read the metadata from the gguf file before loading the model. In the json format the context length is related to the key `llama.context_length`.

abetlen:
@DanieleMorotti does the metadata value differ from `n_ctx_train`? I think this approach is good though, maybe worth changing to the default in a future major release.

DanieleMorotti:
I think ```n_ctx_train``` is the same value that we can read from the metadata. Tell me if I should try to implement the other approach, we would need to add ```gguf-py``` directory of llama.cpp in this repo and call the function that returns the json with the information.
Thanks

Issue #1006, opened 2023-12-13
“<method 'append' of 'list' objects> returned a result with an exception set”


GluttonousCat:
I encountered this error when using the embedding function "llama.create_embedding", under this code. abstract is a text about 700 tokens, when i use the previous version 0.1.83, it can run. now the version is 0.2.22 
```
llama = Llama(model_path='./llama-2-7b.Q4_K_M.gguf', embedding=True, n_ctx=2048, n_gpu_layers=30)
abstract_embedding = llama.create_embedding(input=abstract).get('data')[0].get('embedding')
```
this is the report
```
Traceback (most recent call last):
  File "C:\Users\AI\PycharmProjects\roberta-gat\demo.py", line 107, in <module>
    get_abstract_embedding(path='data/validation.csv', start=0)
  File "C:\Users\AI\PycharmProjects\roberta-gat\demo.py", line 50, in get_abstract_embedding
    abstract_embedding = llama.create_embedding(input=abstract).get('data')[0].get('embedding')
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\AI\.conda\envs\roberta-gat\Lib\site-packages\llama_cpp\llama.py", line 1309, in create_embedding
    data.append(
SystemError: <method 'append' of 'list' objects> returned a result with an exception set
```
This is the source code , and I find the value of llama_cpp.llama_get_embeddings(self._ctx.ctx) maybe is null?
```
for index, input in enumerate(inputs):     tokens = self.tokenize(input.encode("utf-8"), special=True)     self.reset()     self.eval(tokens)     n_tokens = len(tokens)     total_tokens += n_tokens     embedding = llama_cpp.llama_get_embeddings(self._ctx.ctx)[         : llama_cpp.llama_n_embd(self._model.model)     ]      data.append(         {             "object": "embedding",             "embedding": embedding,             "index": index,         }     )
```

CharlieChenDSS:
got the same error with llama-cpp-python v0.2.22 : 

```
  File "/Users/xxxx/opt/anaconda3/envs/langchain_python_v3.9/lib/python3.9/site-packages/llama_cpp/llama.py", line 1309, in create_embedding
    data.append(
SystemError: <method 'append' of 'list' objects> returned a result with an error set

```

CharlieChenDSS:
@abetlen could you please help look into this issue when you get a chance? Thanks. 
[...]

Issue #995, opened 2023-12-11
Add support for new KV Cache Offloading API


abetlen:
Source https://github.com/ggerganov/llama.cpp/pull/4309

Ph0rk0z:
I force on the ctx param to offload kv_cache. The logs print that it's using F16 cache. 81 layers offloaded. memory sizes for K and V. assume KV "should" be. Speeds on 70b are now half and prompt processing is less than 1/4 with no context. Womp Womp.

Ph0rk0z:
Ok, I got it working when I moved the kv to the proper place in the struct.


```
    _fields_ = [
        ("seed", c_uint32),
        ("n_ctx", c_uint32),
        ("n_batch", c_uint32),
        ("n_threads", c_uint32),
        ("n_threads_batch", c_uint32),
        ("rope_scaling_type", c_int8),
        ("rope_freq_base", c_float),
        ("rope_freq_scale", c_float),
        ("yarn_ext_factor", c_float),
        ("yarn_attn_factor", c_float),
        ("yarn_beta_fast", c_float),
        ("yarn_beta_slow", c_float),
        ("yarn_orig_ctx", c_uint32),
        ("type_k", c_int),
        ("type_v", c_int),
        ("mul_mat_q", c_bool),
        ("logits_all", c_bool),
        ("embedding", c_bool),
        ("offload_kqv", c_bool),
    ]
```

PR #1019, opened 2023-12-17
WIP: Complete removal or f16_kv, add offload_kqv field

brandonrobertz:
Adds the `offload_kqv` setting to the context settings fields struct and removes the `f16_kv`. Leaving `f16_kv` there caused a null pointer exception when attempting to use the embeddings example script:

```
$ python examples/high_level_api/high_level_api_embedding.py --model ../../llama/models/llama-2-7b.Q5_K_M.gguf

...
ValueError: NULL pointer access

The above exception was the direct cause of the following exception:

  File "/home/brandon/src/llama-cpp-dev/llama-cpp-python/examples/high_level_api/high_level_api_embedding.py", line 11, in <module>
    print(llm.create_embedding("Hello world!"))
  File "/home/brandon/src/llama-cpp-dev/llama-cpp-python/llama_cpp/llama.py", line 1309, in create_embedding
    data.append(
SystemError: <method 'append' of 'list' objects> returned a result with an exception set
```

This addresses two GH issues:

 - #995 which just requests to add the KV cache offloading param
 - #1006 a NULL ptr exception when using the embeddings (introduced by leaving f16_kv in the fields struct)

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -751,7 +751,6 @@ class Llama:
         yarn_beta_slow: float = 1.0,
         yarn_orig_ctx: int = 0,
         mul_mat_q: bool = True,
-        f16_kv: bool = True,
         logits_all: bool = False,
         embedding: bool = False,
         # Sampling Params
@@ -817,7 +816,6 @@ class Llama:
             yarn_beta_fast: YaRN low correction dim
             yarn_beta_slow: YaRN high correction dim
             yarn_orig_ctx: YaRN original context size
-            f16_kv: Use fp16 for KV cache, fp32 otherwise
             logits_all: Return logits for all tokens, not just the last token. Must be True for completion to return logprobs.
             embedding: Embedding mode only.
             last_n_tokens_size: Maximum number of tokens to keep in the last_n_tokens deque.
@@ -904,7 +902,6 @@ class Llama:
         )
         self.context_params.yarn_orig_ctx = yarn_orig_ctx if yarn_orig_ctx != 0 else 0
         self.context_params.mul_mat_q = mul_mat_q
-        # self.context_params.f16_kv = f16_kv
         self.context_params.logits_all = logits_all
         self.context_params.embedding = embedding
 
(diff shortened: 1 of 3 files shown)


ringohoffman:
I have a question that is unrelated to your PR, which to me does seem to fix the issue in #1006.

What is the significance of the single embedding returned from `llm.create_embedding("Hello world!")`? Since it is composed of 4 tokens, I was expecting 4 embeddings to be returned, but `len(llm.create_embedding("Hello world!")["data"]) == 1`... and the data changes when you provide either `"Hello world!` or `"Hello world"`...

Is `create_embedding` working as expected?

DavidMorton:
I patched this into my current system, and then the embeddings object would just tell me I'd have to create with embedding=True, and it was clearly in the code at the time. Not sure what happened, but I'm wondering if my build of llama_cpp is somehow borked. 

Do older versions work alright as a workaround for this issue? (The one in #1006 )

brandonrobertz:
> I have a question that is unrelated to your PR, which to me does seem to fix the issue in #1006.
> 
> What is the significance of the single embedding returned from `llm.create_embedding("Hello world!")`? Since it is composed of 4 tokens, I was expecting 4 embeddings to be returned, but `len(llm.create_embedding("Hello world!")["data"]) == 1`... and the data changes when you provide either `"Hello world!` or `"Hello world"`...
> 
> Is `create_embedding` working as expected?

It looks like the `create_embedding` method accept two different input types:
- a string, which results in a single embedding
- a list of strings, which results in a list of embeddings (one for each token in the list)

This just mirrors the behavior of the `_LlamaModel.tokenize` method basically. So the behavior doesn't appear to have changed.

brandonrobertz:
> Do older versions work alright as a workaround for this issue? (The one in #1006 )

How old of `llama-cpp-python` are you using? This PR requires relatively new llama-cpp-python and llama.cpp (Dec 11, 2023 or newer basically).

DavidMorton:
image

pip install is 0.2.23 (Dec 13th)

My local branch of llama.cpp's latest commit is slaren's from Dec 16th.

brandonrobertz:
> image
> 
> pip install is 0.2.23 (Dec 13th)
> 
> My local branch of llama.cpp's latest commit is slaren's from Dec 16th.

Should be fine. Can you paste a traceback and the offending code?

DavidMorton:
I managed to get it working. Cleared the pycache and re-ran.  

Thanks!

abetlen:
@brandonrobertz thank you so much for catching this, I'll merge this now and push a new version asap.

PR #1021, opened 2023-12-17
Don't convert logprobs arrays to lists

kddubey:
Follow-up to #1002

## Speedup

Run this in an `ipython` shell (`python -m pip install jupyter` if you don't have it already)

```python
import numpy as np
from llama_cpp import Llama

_scores: np.ndarray = -np.random.default_rng(123).uniform(
    low=0, high=60, size=(20, 32_000)
)
token_offset = 1

print("time old")
%timeit [Llama.logits_to_logprobs(row.tolist()) for row in _scores][token_offset:]
print()
print("time new")
%timeit Llama.logits_to_logprobs(_scores)[token_offset:]

```

prints

```
time old
72.6 ms ± 862 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

time new
24.5 ms ± 142 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

## How has this been tested?

```python
import numpy as np
from llama_cpp import Llama

_scores: np.ndarray = -np.random.default_rng(123).uniform(
    low=0, high=60, size=(20, 32_000)
)
token_offset = 1

# 1-D case
logits = _scores[token_offset, :]
current_logprobs_old = Llama.logits_to_logprobs(logits.tolist())
current_logprobs_new = Llama.logits_to_logprobs(logits)
assert np.allclose(current_logprobs_old, current_logprobs_new)

# 2-D case
all_logprobs_old = [Llama.logits_to_logprobs(row.tolist()) for row in _scores][
    token_offset:
]
all_logprobs_new = Llama.logits_to_logprobs(_scores)[token_offset:]
assert np.allclose(all_logprobs_old, all_logprobs_new)

```

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -1555,7 +1555,7 @@ class Llama:
                             self.detokenize(completion_tokens[:returned_tokens])
                         )
                         token_offset = len(prompt_tokens) + returned_tokens
-                        logits = self._scores[token_offset - 1, :].tolist()
+                        logits = self._scores[token_offset - 1, :]
                         current_logprobs = Llama.logits_to_logprobs(logits)
                         sorted_logprobs = list(
                             sorted(
@@ -1674,7 +1674,7 @@ class Llama:
                         self.detokenize(completion_tokens[:returned_tokens])
                     )
                     token_offset = len(prompt_tokens) + returned_tokens - 1
-                    logits = self._scores[token_offset, :].tolist()
+                    logits = self._scores[token_offset, :]
                     current_logprobs = Llama.logits_to_logprobs(logits)
                     sorted_logprobs = list(
                         sorted(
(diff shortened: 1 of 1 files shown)


llama_cpp/llama.py
```diff
@@ -1788,9 +1788,8 @@ def logit_bias_processor(
                 self.detokenize([token]).decode("utf-8", errors="ignore")
                 for token in all_tokens
             ]
-            all_logprobs = [
-                Llama.logits_to_logprobs(row.tolist()) for row in self._scores
-            ][token_offset:]
+            all_logprobs = Llama.logits_to_logprobs(self._scores)[token_offset:]
```
kddubey, line 1791: `all_logprobs` is now a 2-D array instead of a list of 1-D arrays. But this change should be ok b/c it's zipped, which will iterate over the first axis

abetlen:
@kddubey thank you!

PR #1023, opened 2023-12-17
Update low level example to match current API

jsoma:
Two small changes to make the low level example run successfully:

* Updated call to `llama_cpp.llama_n_vocab` to pass model instead of context
* Combined previous `llama_sample_repetition_penalty` and `llama_sample_frequency_and_presence_penalties` into `llama_sample_repetition_penalties` with keyword arguments

Thank you!


abetlen:
@jsoma thank you

leon-maechler:
something like this is needed again

PR #1028, opened 2023-12-18
Fix ctypes definitions of `llama_kv_cache_view_update` and `llama_kv_cache_view_free`.

e-c-d:
Using the functions llama_kv_cache_view_update and llama_kv_cache_view_free causes a crash. The root cause is a discrepancy between the ctypes definition and the actual definition of the C functions. For example,

```python
# LLAMA_API void llama_kv_cache_view_free(struct llama_kv_cache_view * view);         # <--- This is a pointer to struct
def llama_kv_cache_view_free(view: llama_kv_cache_view):
    """Free a KV cache view. (use only for debugging purposes)"""
    return _lib.llama_kv_cache_view_free(view)


_lib.llama_kv_cache_view_free.argtypes = [llama_kv_cache_view]         # <--- This is the struct itself
```

### DIFF
--- a/llama_cpp/llama_cpp.py
+++ b/llama_cpp/llama_cpp.py
@@ -1041,6 +1041,9 @@ class llama_kv_cache_view(Structure):
     ]
 
 
+llama_kv_cache_view_p = POINTER(llama_kv_cache_view)
+
+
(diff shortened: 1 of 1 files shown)


abetlen:
@e-c-d thank you!

Issue #1051, opened 2023-12-28
LlamaGrammar.from_string should print diagnostics to stderr


turian:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [X] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [X] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [X] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [X] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

llama.cpp should only write the stderr when called from llama-cpp-python.

# Current Behavior

Instead, LlamaGrammar.from_string calls print_grammar to stdout:

https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama_grammar.py?ref=imaurer.com#L75

I expect this should be stderr

PR #1052, opened 2023-12-28
print_grammar to stderr

turian:
Addresses https://github.com/abetlen/llama-cpp-python/issues/1051


abetlen:
@turian thank you for the contribution!

PR #1068, opened 2024-01-06
add ability to pass in penalize_nl param

shankinson:
Let's a user pass in the penalize_nl param to the generate function and passes that to the sample function instead of always using the previous default value of `True`


abetlen:
@shankinson thank you for the contribution!

PR #1078, opened 2024-01-11
Integrate functionary v1.4 and v2 models + add HF AutoTokenizer as optional parameter in llama.create_completion

### LINKED ISSUE #1061: functionary-7b-v1 model upgrade
Suggested model functionary-7b-v1 is quite dumb, v1.4 uses Mistral, and v2.X is using new OpenAI API. Please support those new https://github.com/MeetKai/functionary models.
When I've try to use llama-cpp-python as an OpenAI api substitution with v1.4 I've got error:

`Something went wrong: function ' execute_services' does not exist` - like extra space within function name, when I use yaml as function configuration then sometimes it is ': function_name'

This is related to llama-cpp-python or functionary maybe ?

### LINKED ISSUE #1075: Refactor Functionary chat_handler to use HF AutoTokenizer
Currently, llama-cpp's tokenizer has two differences/problems as compared to HF's AutoTokenizer.
1. `model.tokenize` (llama-cpp-python) != `tokenizer.encode` (HF)
llama-cpp's tokenizer seems to always generate a different token from HF's autotokenizer for the token right after every special token.
2. model.detokenize returns empty string when trying to convert a special token back to text

I am currently working on implementing the chat handlers for [Functionary-7b-v1.4 and all v2 models](https://huggingface.co/meetkai) as mentioned in [this issue](https://github.com/abetlen/llama-cpp-python/issues/1061). However, all the models use added special tokens in the prompt template and as stopping tokens. This results in suboptimal generation (due to problem 1) and the inability to stop generation (problem 2).

Refactor the current Functionary chat_handler to use HF AutoTokenizer instead of llama-cpp-python's tokenizer. We can then directly use the jinja chat template in the various Functionary models. This also avoids any discrepancies in terms of tokenization.

**Describe alternatives you've considered**
Fixing the tokenizer issue in llama.cpp directly but lack the knowledge of the details of that codebase.

**Additional context**
NA

jeffrey-fong:
Closes #1075 
Closes #1061 

- added Optional parameter to allow tokenizing using HF AutoTokenizer in functionary chat_completion_handler and llama.create_completion
- Replaced the logic for the obsolete functionary-7b-v1 model in functionary_chat_handler with both functionary v1.4 and v2.* models
- Implemented parallel function calling feature as v2.* models support it

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,12 +329,12 @@ Note that due to discrepancies between llama.cpp and HuggingFace's tokenizers, i
           }
         }
       }],
-      tool_choice=[{
+      tool_choice={
         "type": "function",
         "function": {
           "name": "UserDetail"
         }
-      }],
+      },
 )
 ```
 
--- a/README.md
+++ b/README.md
@@ -301,7 +301,7 @@ Note that due to discrepancies between llama.cpp and HuggingFace's tokenizers, i
 >>> from llama_cpp import Llama
 >>> llm = Llama(model_path="path/to/functionary/llama-model.gguf", chat_format="functionary")
 >>> llm.create_chat_completion(
-      hf_tokenizer_path="path/to/functionary-gguf/"
+      hf_tokenizer_path="path/to/functionary-gguf/",
       messages = [
         {
           "role": "user",
--- a/README.md
+++ b/README.md
@@ -293,15 +293,14 @@ To constrain the response to a specific JSON Schema, you can use the `schema` pr
 
 The high-level API also provides a simple interface for function calling.
 
-The only set of models that supports full function calling at this time is [functionary](https://github.com/MeetKai/functionary). The various gguf-converted files for this set of models can be found [here](https://huggingface.co/meetkai). Functionary is able to intelligently call functions and also analyze any provided function outputs to generate coherent responses. All v2 models of functionary supports **parallel function calling**.
+The only set of models that supports full function calling at this time is [functionary](https://github.com/MeetKai/functionary). The various gguf-converted files for this set of models can be found [here](https://huggingface.co/meetkai). Functionary is able to intelligently call functions and also analyze any provided function outputs to generate coherent responses. All v2 models of functionary supports **parallel function calling**. You can provide either `functionary-v1` or `functionary-v2` for the `chat_format` when initializing the Llama class.
 
 Note that due to discrepancies between llama.cpp and HuggingFace's tokenizers, it is required to provide the path to the HF tokenizer for functionary. They are already included in the respective HF repositories hosting the gguf files.
 
 ```python
 >>> from llama_cpp import Llama
->>> llm = Llama(model_path="path/to/functionary/llama-model.gguf", chat_format="functionary")
+>>> llm = Llama(model_path="path/to/functionary/llama-model.gguf", hf_tokenizer_path="path/to/functionary-gguf/", chat_format="functionary-v2")
 >>> llm.create_chat_completion(
-      hf_tokenizer_path="path/to/functionary-gguf/",
       messages = [
         {
           "role": "user",
--- a/README.md
+++ b/README.md
@@ -293,20 +293,16 @@ To constrain the response to a specific JSON Schema, you can use the `schema` pr
 
 The high-level API also provides a simple interface for function calling.
 
-Note that the only set of models that supports full function calling at this time is [functionary](https://github.com/MeetKai/functionary).
-The various gguf-converted files for this set of models can be found [here](https://huggingface.co/meetkai).
+The only set of models that supports full function calling at this time is [functionary](https://github.com/MeetKai/functionary). The various gguf-converted files for this set of models can be found [here](https://huggingface.co/meetkai). Functionary is able to intelligently call functions and also analyze any provided function outputs to generate coherent responses. All v2 models of functionary supports **parallel function calling**.
+
 Note that due to discrepancies between llama.cpp and HuggingFace's tokenizers, it is required to provide the path to the HF tokenizer for functionary. They are already included in the respective HF repositories hosting the gguf files.
 
 ```python
 >>> from llama_cpp import Llama
 >>> llm = Llama(model_path="path/to/functionary/llama-model.gguf", chat_format="functionary")
 >>> llm.create_chat_completion(
+      hf_tokenizer_path="path/to/functionary-gguf/"
       messages = [
-        {
-          "role": "system",
-          "content": "A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful, detailed, and polite answers to the user's questions. The assistant calls functions with appropriate input when necessary"
-
-        },
         {
           "role": "user",
           "content": "Extract Jason is 25 years old"
@@ -339,7 +335,6 @@ Note that due to discrepancies between llama.cpp and HuggingFace's tokenizers, i
           "name": "UserDetail"
         }
       }],
-      hf_tokenizer_path="path/to/functionary-gguf/"
 )
 ```
 
--- a/README.md
+++ b/README.md
@@ -12,19 +12,16 @@ This package provides:
 
 - Low-level access to C API via `ctypes` interface.
 - High-level Python API for text completion
-    - OpenAI-like API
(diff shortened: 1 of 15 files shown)


sandangel:
This is great. I'm waiting for this PR. can we merge this soon? @abetlen ?

llama_cpp/llama.py
```diff
@@ -1369,11 +1370,24 @@ def _create_completion(
         logits_processor: Optional[LogitsProcessorList] = None,
         grammar: Optional[LlamaGrammar] = None,
         logit_bias: Optional[Dict[str, float]] = None,
+        hf_tokenizer: Optional[Any] = None,
```
abetlen, line 1373: Can we introduce a `typing.Protocol` for a tokenizer and move this up to the `Llama.__init__` as an optional override to the default tokenizer.

llama_cpp/llama_chat_format.py
```diff
@@ -783,9 +785,30 @@ def functionary_chat_handler(
     model: Optional[str] = None,
     logits_processor: Optional[llama.LogitsProcessorList] = None,
     grammar: Optional[llama.LlamaGrammar] = None,
+    hf_tokenizer_path: Optional[str] = None,
     **kwargs,  # type: ignore
 ) -> Union[llama_types.ChatCompletion, Iterator[llama_types.ChatCompletionChunk]]:
     SYSTEM_MESSAGE = """A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful, detailed, and polite answers to the user's questions. The assistant calls functions with appropriate input when necessary"""
+
```
abetlen, line 792: I know the original functionary model is deprecated but I want to breaking any code that's using the old model, can we introduce these chat formats as `functionary-v1` and `funtionary-v2` for the new models.

abetlen (changes requested):

rgbkrk:
I'm going to start using this branch directly. Thanks for putting it together!

README.md
```diff
@@ -198,20 +198,17 @@ Chat completion is available through the [`create_chat_completion`](https://llam
 
 The high-level API also provides a simple interface for function calling.
 
-Note that the only model that supports full function calling at this time is "functionary".
-The gguf-converted files for this model can be found here: [functionary-7b-v1](https://huggingface.co/abetlen/functionary-7b-v1-GGUF)
+The only set of models that supports full function calling at this time is [functionary](https://github.com/MeetKai/functionary). The various gguf-converted files for this set of models can be found [here](https://huggingface.co/meetkai). Functionary is able to intelligently call functions and also analyze any provided function outputs to generate coherent responses. All v2 models of functionary supports **parallel function calling**.
+
+Note that due to discrepancies between llama.cpp and HuggingFace's tokenizers, it is required to provide the path to the HF tokenizer for functionary. They are already included in the respective HF repositories hosting the gguf files.
```
rgbkrk, line 203: What do you think about using `hf_hub_download` to get the model and to acquire the local file path?

```python
from huggingface_hub import hf_hub_download

# Model repository on the Hugging Face model hub
model_repo = "meetkai/functionary-small-v2.2-GGUF"

# File to download
file_name = "functionary-small-v2.2.f16.gguf"

# Download the file
local_file_path = hf_hub_download(repo_id=model_repo, filename=file_name)
local_file_path
```

rgbkrk:
On this branch I get a segfault on launch. On `main` I don't. This is what I run:

```
python -m llama_cpp.server --model $(python -c "from huggingface_hub import hf_hub_download; print(hf_hub_download(repo_id='meetkai/functionary-small-v2.2-GGUF', filename='functionary-small-v2.2.f16.gguf'))")  --chat_format functionary
```

If I rebase this branch against `main` it starts fine then hits the AssertionError for providing a valid hf tokenizer path.

llama_cpp/llama_chat_format.py
sandangel: I think the code already checked the model version to use the proper chat template

jeffrey-fong: I see that you have a branch that keeps the old v1 code. I will just copy those relevant parts from there into my branch and deduplicate as much code as possible between the new and old chat handlers

README.md
jeffrey-fong: Thanks for the good suggestion! However, I think it will be difficult to download the model with huggingface_hub as huggingface_hub is not a dependency in this project. I'm not sure if @abetlen is ok with setting this as a dependency?

It's kinda pointless if we set up `hf_hub_download` just for the HF AutoTokenizer since we would have already git cloned the repository to get the model.

jeffrey-fong:
@abetlen I've made the changes. How does this look now? Mainly tokenizer was changed to adopt an ABC-ish setup with `typing.Protocol` as suggested by you. The documentation page for the OpenAI server may need to be updated since I've added `hf_tokenizer_path` into `ModelSettings` too.

README.md
rgbkrk: Good to know. `hf_hub_download` does cache the path so for any user that already has it (via transformers, etc.), it won't redownload it.

sandangel:
Hi, may I ask if there is updates on this PR? I think the author has implemented the requested change. Can we merge? 😊

abetlen:
@jeffrey-fong looks good, I'm not able to modify the PR but I do have a couple of changes I still need to make, I'll just make them after the merge:

- Instead of `hf_tokenizer_path` in the `__init__` I'm going to change this to an optional `tokenizer` param so you can pass an `HFTokenizer` to the `Llama` class. I'll keep the `hf_tokenizer_path` option for the server though.
- I'm going to revert the tokenizer signature and adjust the surrounding code otherwise this might break some existing code expecting the old signature.

Additionally I'm going to create an issue so that we can fix the `llama.cpp` tokenizer problems that are preventing us from using the normal tokenizer as-is.

jeffrey-fong:
@abetlen Sure, sounds good to me! I'll keep track of the issue too as it feels really inelegant to have to pass in a HF Tokenizer.

I've pulled from main so feel free to approve and merge it. 👍

abetlen:
Hey @jeffrey-fong still working on the changes over in my `origin/integrate-functionary` branch, not sure if I screwed something up but when streaming normally the tokenizer seems to ignore any spaces.

image

Just wondering if you've seen that before.

jeffrey-fong:
@abetlen seems like it's due to this line in llama.py https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama.py#L937

The Llama HF Tokenizer does not prepend the prefix space to the first token in the token list when decoding as mentioned here:
image
https://huggingface.co/docs/transformers/main/model_doc/llama

I have modified the streaming detokenization in the `_create_completion` method with offset mapping logic when the tokenizer is LlamaHFTokenizer. Hope it does not break any existing code.

abetlen:
@jeffrey-fong is it possible to account for this in the HFTokenizer instead of `_create_completion`?

jeffrey-fong:
@abetlen sure, I've moved this to both HFTokenizer and LlamaTokenizer. `Llama.detokenize` will have a new Optional parameter called `prev_tokens`. If `prev_tokens is not None`, we will perform offset mapping in the decoded string level for HFTokenizer and tokens level for LlamaTokenizer. Hope this new optional parameter is ok for you.

abetlen:
@jeffrey-fong thanks for your patience with this PR, I'll review again tonight.

Also, in my #957 I was able to write up some helpers to add support for streaming function calls so I should be able to integrate that for functionary as well once it's merged.

sandangel:
﻿Hello, may I inquire if there's an update on this issue? It appears we're awaiting a PR merge. Am I correct in understanding that?

abetlen (approved):

abetlen:
Sorry I've just been getting over the flu last few days didn't have a chance to work on this. Looks good, I'll merge and I may still update the tokenizer api before the next release but it should be in and working!

Thanks so much for the work on this @jeffrey-fong

rgbkrk:
Hope you rested up well @abetlen. Thank you!

sandangel:
Please rest well, @abetlen. Thank you for working on this. @jeffrey-fong @abetlen

PR #1086, opened 2024-01-14
Fix low_level_api_chat_cpp example to match current API

aniljava:


abetlen:
@aniljava thank you for fixing the example! Really appreciate it.

cbigger:
I'm not sure if this is entirely updated. When running the Chat.py example from the low level api example I get the following error when it gets to generation:

`AttributeError: module 'llama_cpp' has no attribute 'llama_eval'`

llama_eval is supposedly deprecated when i trace back to llama_cpp:

`logits_all (bool): the llama_eval() call computes all logits, not just the last one (DEPRECATED - set llama_batch.logits instead)`

The examples I saw referenced ggml.bin models as well - not sure if that is intentional or not.

Issue #1048, opened 2023-12-27
Can't create grammar from JSON schema if referencing multiple definitions


DavideWalder:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [x] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [x] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

I can't create a `LlamaGrammar` from a JSON schema if multiple definitions are referenced. See example below.
It might also be the root cause of https://github.com/abetlen/llama-cpp-python/issues/1043.

The feature should either be implemented or an explicit error should be raised.

# Current Behavior

Throws an error since refs are not supported

# Environment and Context

```
❯ uname -a
Darwin Davides-MacBook-Pro-2.local 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:55:06 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6020 arm64
```
# Failure Information (for bugs)

```
AssertionError: Unrecognized schema: {'$ref': '#/definitions/A'}
```

# Steps to Reproduce

```python
from llama_cpp.llama_grammar import LlamaGrammar
from pydantic import BaseModel


class A(BaseModel):
    some_field: str


class B(BaseModel):
    a: A


LlamaGrammar.from_json_schema(B.schema_json())
```

nyanp:
I encountered the same issue. One workaround I found is to inlining references using [jsonref](https://github.com/gazpachoking/jsonref):

```python
import json
from llama_cpp.llama_grammar import LlamaGrammar
from pydantic import BaseModel
from jsonref import replace_refs


class A(BaseModel):
    some_field: str


class B(BaseModel):
    a: A


def json_schema_with_inlining(model):
    replaced = replace_refs(model.schema(), proxies=False)
    if "$defs" in replaced:
        del replaced["$defs"]
    return json.dumps(replaced)


grammar = LlamaGrammar.from_json_schema(json_schema_with_inlining(B))
```

[...]

Issue #1043, opened 2023-12-24
OpenAI Compatible server (Function calling), does not support (jxnl/Instructor), because of Unrecognized schema: {'$ref': '#/$defs/Edge'}


PAHXO:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [OK ] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [OK ] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [ OK] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [OK] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

I am currently using `jxnl/Instructor`, a package that patches OpenAI API to add extra features related to function calling and generating structured data.

While using it on the official OpenAI function calling API. it works very well.

When I use the included OpenAI compatible server with Instructor. it works for simple structured data, but as soon as I try to use anything more complex. it throws an error and says and unrecognized schema.

# Current Behavior
Currently it works with simple data when I use instructor.

but as soon as I try to use something more complex like generating Knowledge graphs.
it does not work, and throws an error then uses the default grammar.
sometimes it could spit out a correct output (with the default grammar), but mostly does not work.

# Environment and Context

Windows, python 3.10 venv.

pip install llama-cpp-python[server]
python3 -m llama_cpp.server --model <model_path>

This code works, Code from https://github.com/jxnl/instructor?tab=readme-ov-file#usage
```python
import instructor
from openai import OpenAI
from pydantic import BaseModel

# Enables `response_model`
client = instructor.patch(OpenAI(api_key="sk-****",base_url="https://my-endpoint/v1"))

class UserDetail(BaseModel):
    name: str
    age: int

user = client.chat.completions.create(
    model="gpt-3.5-turbo",
    response_model=UserDetail,
    messages=[
        {"role": "user", "content": "Extract Jason is 25 years old"},
    ]
)

assert isinstance(user, UserDetail)
assert user.name == "Jason"
assert user.age == 25
```

---

This does not work, from https://jxnl.github.io/instructor/examples/knowledge_graph/#defining-the-structures .
```python
[...]

PR #1087, opened 2024-01-14
fix Pydantic model parsing

DeNeutoy:
Fixes https://github.com/abetlen/llama-cpp-python/issues/1048, https://github.com/abetlen/llama-cpp-python/issues/1043

### Problem

When parsing json schemas generated from a Pydantic model, if the schema has references to other models, it contains a `$defs` key, which are referred to in the schema itself via `$ref`. This `$defs` key does not have a `type` key, as it is actually defining the types - and so the `SchemaConverter` throws an error, as it requires *all* schema keys to have a type. This assert is actually not required, as it is caught later on after iterating through the various options for schema_types here:

https://github.com/abetlen/llama-cpp-python/compare/main...DeNeutoy:fix-model-parsing?expand=1#diff-22ae5683f6e444008ca0411a78851823be188bd8e7f287327397462c5671a1a4L1502

I've deleted it and added a test. Cool project, thanks!

### DIFF
--- a/llama_cpp/llama_grammar.py
+++ b/llama_cpp/llama_grammar.py
@@ -1433,7 +1433,6 @@ class SchemaConverter:
 
     def visit(self, schema: Dict[str, Any], name: str) -> str:
         schema_type: Optional[str] = schema.get("type") # type: ignore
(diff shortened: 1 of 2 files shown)


abetlen:
@DeNeutoy thank you for the contribution!

PR #1090, opened 2024-01-16
Add self extend support

Josh-XT:
Add self extend support
- More info in https://github.com/ggerganov/llama.cpp/pull/4815

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -751,6 +751,8 @@ class Llama:
         yarn_beta_fast: float = 32.0,
         yarn_beta_slow: float = 1.0,
         yarn_orig_ctx: int = 0,
+        grp_attn_n: int = 1,
+        grp_attn_w: int = 512,
         mul_mat_q: bool = True,
         logits_all: bool = False,
         embedding: bool = False,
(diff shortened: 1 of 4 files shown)


Josh-XT:
I might still be missing something, still checking.

Josh-XT:
Haven't had time to finish the rest of this, got pulled away in the middle! Sorry - I'll reopen when ready.

jshph:
Hey @Josh-XT thanks for taking a stab. Curious if the implementation is in the right direction here, I was poking at it yesterday and saw the Params struct being modified. I wonder if that'd affect the parsing of args downstream.

Josh-XT:
> Hey @Josh-XT thanks for taking a stab. Curious if the implementation is in the right direction here, I was poking at it yesterday and saw the Params struct being modified. I wonder if that'd affect the parsing of args downstream.

I was attempting to follow the changes made on the cpp repo but while hanging out with my 4 year old, then he made a mess and I had to stop. 🤣 I was trying to do the cpp server example at the same time when my plans were changed. The implementation is pretty straightforward. I think the `grp-attn-n` and `grp-attn-w` could (and probably will) be automated away.

ExtReMLapin:
Any update on this ?

PR #1116, opened 2024-01-22
GGUF metadata KV overrides, re #1011

phiharri:
(starting from a clean branch)

76aafa6 didn't work for me, `_kv_overrides_array[i]` refs uninitialised `i` and I think it's necessary to instantiate the CStructs array with `()` or it ends up a generic PyCArrayType.

Let me know what you think about this PR instead.

Simple validation that it takes effect:
```
>>> llm=llama_cpp.Llama(model_path="../models/mixtral-8x7b-instruct-v0.1.Q6_K.gguf", kv_overrides={"llama.embedding_length":4095})

validate_override: Using metadata override (  int) 'llama.embedding_length' = 4095
llama_model_load: error loading model: invalid n_rot: 128, expected 127 
```

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -216,6 +216,7 @@ class Llama:
                 else:
                     raise ValueError(f"Unknown value type for {k}: {v}")
 
+            self._kv_overrides_array[-1].key = b'\0'  # ensure sentinel element is zeroed
             self.model_params.kv_overrides = self._kv_overrides_array
 
         self.n_batch = min(n_ctx, n_batch)  # ???
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -194,33 +194,31 @@ class Llama:
         self.model_params.use_mmap = use_mmap if lora_path is None else False
         self.model_params.use_mlock = use_mlock
 
+        # kv_overrides is the original python dict
+        # _kv_overrides_array is a list of llama_model_kv_override CStructs
         self.kv_overrides = kv_overrides
+        self._kv_overrides_array = []
         if kv_overrides is not None:
-            n_overrides = len(kv_overrides)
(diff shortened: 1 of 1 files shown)


abetlen:
@phiharri looks good, is the sentinel value not required, I saw that in your original PR.

phiharri:
Not sure where I got that idea from tbh, pretty sure it isn't needed. Thought about checking len(key)<128 for potential overflow but it seems ctypes takes care of that already 🤷

```
>>> k='a'*130
>>> kvo.key=k.encode('utf-8')
  File "<stdin>", line 1, in <module>
ValueError: bytes too long (130, maximum length 128)
```

abetlen:
I think it is necessary to add an additional key and set it to the null pointer, see https://github.com/ggerganov/llama.cpp/blob/6f9939d119b2d004c264952eb510bd106455531e/llama.cpp#L2250

phiharri:
Sure, how's this look? Elements get initialised empty/null but if you wanted to be explicit we can add
`self._kv_overrides_array[-1].key = b'\0'`

abetlen:
@phiharri that's correct however I think(?) we still need to store the reference to that byte string because we're passing it into a c data structure and want to avoid any memory issues.

phiharri:
Aren't we just modifying data in the existing `_kv_overrides_array`, not allocating new memory (unless doing create_string_buffer)?

Definitely seems the array gets null initialised so I don't think `self._kv_overrides_array[-1].key = b'\0'` is needed. https://docs.python.org/3/library/ctypes.html#arrays  "the array contents is initialized to zeros"

Peeking the memory of the last element without doing any `b'\0'` assignment shows all zeros:

null init

abetlen:
Doh you're right, I thought the key was a `char *` not a fixed size array, yes as long as it's zero initialized (I'd like to be explicit about this) then it should be okay.

abetlen:
Awesome, okay good to merge it then. Thanks for the contribution @phiharri

PR #1120, opened 2024-01-23
Add speculative decoding support

abetlen:
Uses prompt lookup decoding but the draft model class can be extended to support almost any existing method.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -41,6 +41,8 @@ from ._internals import (
     _LlamaContext,  # type: ignore
     _LlamaBatch,  # type: ignore
     _LlamaTokenDataArray,  # type: ignore
+    _LlamaSamplingParams, # type: ignore
+    _LlamaSamplingContext, # type: ignore
 )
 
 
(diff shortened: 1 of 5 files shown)


abetlen:
Tried on a more realistic example and got worse performance, think I'll need to tune / implement a heuristic for draft models similar to https://huggingface.co/blog/assisted-generation

> Adjust the number of candidate tokens to be produced in the next iteration — our original heuristic increases it by 2 if ALL tokens match and decreases it by 1 otherwise.

abetlen:
Added the adaptive heuristic and it does do better but still occasionally slower even with termperature=0, will need to investigate.

oobabooga:
Highly appreciated PR. Is it possible to make `prompt_lookup_num_tokens` a generation parameter on the same footing as temperature as is done in the transformers library? That would make it possible to change that parameter without having to reload the model.

abetlen:
@oobabooga I saw, I was looking at the hf implementation as a reference. I could add it as a general `n_candidate_tokens` because I want to keep it open to other implementations of speculative decoding. I'll think on that one.

abetlen:
@oobabooga going to merge this now. For updating the draft model or it's properties without re-creating the entire `Llama` model class just assume that you can access `llm.draft_model`, set it to `None` to disable.

oobabooga:
Awesome, thanks @abetlen!

PR #1131, opened 2024-01-26
Added functionary v2.2 chat handler

AbdoMahfoz:
- Supports streaming and no streaming
- Tested on local project
- I used the same code that generated schema for tool calls in `functionary` chat handler, rewrote almost everything else as they completely changed the format.
- Built-in tokenizer had problems tokenizing special tokens like `<|content|>` and `<|stop|>` so I had to use `transformers` package to get an `AutoTokenizer` that works with this model and just plug it with `llama.generate`.
  - As a side effect of this, `sentencepiece` is now a requirement.
- Updated readme to reflect support for the newest model the I tested as of now.
- Chat format is set as `functionary2` to retain compatibility with any code that uses the old formatter with older models.

Here are samples of the [new chat format](https://github.com/MeetKai/functionary/blob/main/tests/prompt_test_v2.txt) and [old chat format](https://github.com/MeetKai/functionary/blob/main/tests/prompt_test_v1.txt).

Please note that current `functionary` handler **didn't work** with functionary model v1.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,12 +221,12 @@ Chat completion is available through the [`create_chat_completion`](https://llam
 The high-level API also provides a simple interface for function calling.
 
 Note that the only model that supports full function calling at this time is "functionary".
-The gguf-converted files for this model can be found here: [functionary-7b-v1](https://huggingface.co/abetlen/functionary-7b-v1-GGUF)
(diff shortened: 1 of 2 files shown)


abetlen:
@AbdoMahfoz thank you so much for the PR and sorry for responding so late, at the time jeffrey from the functionary team had just submitted the same work shortyl before you so I merged that one. Thank you again however.

AbdoMahfoz:
No worries ❤️. I needed it in a local project so it served me well all that time 😂

PR #1136, opened 2024-01-27
Add endpoint to count tokens

felipelo:


abetlen:
Hey @felipelo , my thought was to add `/tokenize` and `/detokenize` endpoints to the server, happy to include just an endpoint for counting but let's include it as `/tokenize/count` or something along those lines.

lucas-bortoli:
Hey, can we look into this again? I'm using llama-cpp-python's webserver for an application I'm building and it would be nice to have the tokenizer available as an API. This looks great!

abetlen:
Sorry for the delay, just had to do some cleanup, this will be in the next version.

PR #1170, opened 2024-02-09
feat: implement required attributes in json_schema_to_gbnf

BDav24:
If a JSON schema include required attributes, the optional attributes should have `| "null"` in the generated grammar.

Example:
```python
schema = {
    "properties": {
        "a": {"type": "integer"},
        "b": {"type": "integer"},
    },
    "required": ["a"],
    "type": "object",
}

llama_cpp.llama_grammar.json_schema_to_gbnf(json.dumps(schema))
```

Should have the following grammar:
```
space ::= " "?
integer ::= ("-"? ([0-9] | [1-9] [0-9]*)) space
integer-or-null ::= (("-"? ([0-9] | [1-9] [0-9]*)) | "null") space
root ::= "{" space "\"a\"" space ":" space integer "," space "\"b\"" space ":" space integer-or-null "}" space
```


abetlen:
@BDav24 I tried that example with the `jsonschema` package and it doesn't seem that `null` is valid as a stand-in for non-required fields.

BDav24:
@abetlen Yes on one hand you're completely right:

```python
from jsonschema import validate

schema = {
    "properties": {
        "a": {"type": "integer"},
        "b": {"type": "integer"},
    },
    "required": ["a"],
    "type": "object",
}

object = {"a": 1, "b": None}

validate(instance=object, schema=schema)
# jsonschema.exceptions.ValidationError: None is not of type 'integer'
```

I made it this way for practical reasons: when I ask a model to extract information, I prefer to have all the keys in the result to be sure that my request was "understood". Though the right schema would contain `{"anyOf": [{"type": "..."}, {"type": "null"}]}`, I find it too verbose to be put in a prompt.

We could always remove empty keys to validate the object:

```python
validate(instance={k: v for k, v in object.items() if v}, schema=schema)
# => OK
```

In conclusion, I think this is pratical but I agree it's not the cleanest, I'll let you decide if the PR needs to be changed.

BDav24:
@abetlen I implemented it so that the output is compatible with `jsonschema`, but also added an option `treat_optional_as_nullable` to allow nullable properties.

Note: if you have a schema with a lot of optional properties (and no required property at the beginning), then the grammar will have a lot of permutations, but that's the only I found to handle the absence of the last comma.

PR #1172, opened 2024-02-10
[README] Set the correct command for compiling with sycl support

akarshanbiswas:
If incorrect compiler is set (such as `/usr/bin/c++`), the compilation will fail. This change proposes the correct steps to compile with sycl support on Intel Arc GPUs.


abetlen:
@akarshanbiswas thank you!

PR #1176, opened 2024-02-11
Allow Llama objects to be freed earlier again

notwa:
commit 901827013b732d74f1f67033062d13a6204a62bd introduced a cyclic dependency within Llama objects. That change causes old models to linger in memory longer than necessary, thereby creating memory bloat in most applications attempting to switch between models at runtime. This patch simply removes the problematic line, allowing models to deallocate without relying on GC. One might also consider combining `weakref.ref` with a `@property` if the `llama` attribute is absolutely necessary to expose in the tokenizer class.


abetlen:
@notwa good catch!

PR #1179, opened 2024-02-12
fix sample_idx off-by-one error

lapp0:
Fixes an off-by-one error when applying logits processor.

In this example, a draft model isn't used. The `sample_idx` is set to one less than the length of the input tokens, resulting in the last token being truncated.

```
(Pdb) idx
8
(Pdb) self._input_ids
array([    1,  1824,   349,   272,  1141,   302,  5459,  9955, 28804],
      dtype=int32)
(Pdb) len(self._input_ids)
9
(Pdb) self._input_ids[:idx]
array([   1, 1824,  349,  272, 1141,  302, 5459, 9955], dtype=int32)
```

@abetlen Would appreciate a review since you have a better understanding of the `sample_idx` argument. Especially considering this will impact `logits = self._scores[idx, :]`

Related: https://github.com/outlines-dev/outlines/issues/638


abetlen:
@lapp0 I'll look into this, is it specifically an issue with the `logits_processor`?

lapp0:
@abetlen yes it is.

abetlen:
@lapp0 thank you, looks good to me!

lapp0:
Thanks so much!

PR #1186, opened 2024-02-13
Batch embed strings to fit model context

iamlemec:
Now that `llama.cpp` supports BERT embedding models (https://github.com/ggerganov/llama.cpp/pull/5423), this modifies `embed` / `create_embedding` to truncate and combine inputs to fit into the model context size and allows for efficient high-volume embedding. I tried to be minimal with changes, and am definitely open to alternate suggestions. What we have now:

* Add `reset` and `add_sequence` methods to `_LlamaBatch` to make multi-sequence batches
* Tokens are truncated to `n_ctx` by default, and without truncation an error is thrown if `n_tokens > n_ctx`
* Embeddings are normalized by default but this can be turned off

One could eke out some gains from grouping batches more intelligently, but I wanted to keep it simple.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -717,9 +717,43 @@ class Llama:
         Returns:
             An embedding object.
         """
-        assert self._ctx.ctx is not None
         assert self._model.model is not None
         model_name: str = model if model is not None else self.model_path
+
+        # get numeric embeddings
+        embeds, total_tokens = self.embed(input, return_count=True)
+
+        # convert to CreateEmbeddingResponse
+        data = [
+            {
+                "object": "embedding",
+                "embedding": emb,
+                "index": idx,
+            } for idx, emb in enumerate(embeds)
+        ]
+
+        return {
+            "object": "list",
+            "data": data,
+            "model": model_name,
+            "usage": {
+                "prompt_tokens": total_tokens,
+                "total_tokens": total_tokens,
+            },
+        }
+
+    def embed(self, input: str, normalize: bool = True, truncate: bool = True, return_count: bool = False) -> List[float]:
+        """Embed a string.
+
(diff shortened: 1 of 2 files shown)


abetlen:
@iamlemec absolute legend, thank you!

I'll test that this doesn't interfere with the current prefix-matching kv logic and once that looks good I should be able to merge. Wrt normalizing the embeddings what do you think are the benefits for / against?

iamlemec:
I could go either way on the default for `normalize`. It looks like `sentence_transformers` defaults to `normalize_embeddings=False`. But actually, their results come back the same regardless of what you set this flag to, so I guess it's getting normalized anyway somewhere else?

From a practical perspective, I assume most people are using these for cosine similarities, in which case you'd want to normalize them. But normalization is also a destructive process, so in some sense not normalizing is the safer bet.

Also, I should note that the the `llama.cpp` code returns the pooled sum over token embeddings, so we also divide here by the number of tokens to make it a pooled mean, which is more standard.

llama_cpp/llama.py
```diff
@@ -734,48 +777,72 @@ def create_embedding(
...
+                )
+
+            # time to eval batch
+            if t_batch + n_tokens > self._n_ctx:
+                decode_batch(s_sizes)
+                t_batch = 0
+                s_sizes = []
+
+            # add to batch
+            self._batch.add_sequence(tokens, len(s_sizes), False)
```
cebtenzzre, line 827: @iamlemec There is a null pointer access in this function if `n_batch < n_ctx` and the prompt exceeds n_batch.

iamlemec: Yup! Check out #1194. It changes the bounds checking from `n_ctx` to `n_batch`.

PR #1188, opened 2024-02-14
switch to '\n' for EventSourceResponse separator

khimaros:
this fixes compatibility with some OpenAI clients, including BetterChatGPT (https://github.com/ztjhz/BetterChatGPT/issues/537). i'm not sure if this will cause problems with other clients. i'm also working with the client to see if they are willing to handle these separators more gracefully.

### DIFF
--- a/llama_cpp/server/app.py
+++ b/llama_cpp/server/app.py
@@ -290,6 +290,7 @@ async def create_completion(
                 inner_send_chan=send_chan,
                 iterator=iterator(),
             ),
+            sep='\n',
         )
     else:
         return iterator_or_completion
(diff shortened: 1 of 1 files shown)


khimaros:
note that BetterChatGPT client still has trouble with `sse-starlet` ping messages as described in https://github.com/ztjhz/BetterChatGPT/issues/538 -- the hacky workaround i'm using locally for this is to set the `ping=100000` on the `EventSourceResponse` -- sse-starlet doesn't seem to provide a way to disable these entirely so it seems like a client side solution would be ideal.

abetlen:
@khimaros makes sense, I'm assuming the `\n` seperator is what's used by the openai api when doing server sent events / streaming.

abetlen:
Thank you, I'll go ahead and merge this.

PR #1194, opened 2024-02-15
Incorporate embedding pooling layer fixes

iamlemec:
Made some fixes to the pooling layer in `llama.cpp` that are reflected here. Previously we had to divide by the number of tokens in the sequence. Now we can just take them as-is and optionally normalize. Also changed the truncation to `n_batch` rather than `n_ctx` since that's what we're writing to.

Embedding numbers now match up very closely with SentenceTransformers. Usually around `1-(1e-7)` cosine similarity, though there are some remaining issues with tokenizing text with accents.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -762,7 +762,7 @@ class Llama:
         """
         assert self._ctx.ctx is not None
         n_embd = self.n_embd()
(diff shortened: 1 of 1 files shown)


abetlen:
Thanks @iamlemec

PR #1210, opened 2024-02-22
Add Google's Gemma formatting via `chat_format="gemma"`

alvarobartt:
## Description

This PR adds the recently released Google's Gemma formatting, so that it can be used via the `chat_format` arg within the `Llama` class i.e. `chat_format="gemma"`.

More information about the models in the following HuggingFace Collection at https://huggingface.co/collections/google/gemma-release-65d5efbccdbb8c4202ec078b

_Note that some of the current GGUF versions uploaded to the HuggingFace Hub are not working fine with `llama.cpp`, but I got it working when using e.g. https://huggingface.co/rahuldshetty/gemma-7b-it-gguf-quantized/blob/main/gemma-7b-it-Q4_K_M.gguf._

## Example

```python
from llama_cpp import Llama

llm = Llama(
    model_path="./models/gemma-7b-it-Q4_K_M.gguf",
    chat_format="gemma",
    n_gpu_layers=-1,
)
print(
    llm.create_chat_completion(
        messages=[
            {"role": "user", "content": "What's the capital of Spain"},
            {"role": "assistant", "content": "Barcelona"},
            {"role": "user", "content": "No, it's not, try again."},
        ],
        max_tokens=4,
    )
)
# {'id': 'chatcmpl-6037698e-2f05-496d-b49f-f7d61dde32f3', 'object': 'chat.completion', 'created': 1708599453, 'model': './models/gemma-7b-it-Q4_K_M.gguf', 'choices': [{'index': 0, 'message': {'role': 'assistant', 'content': 'The answer is Madrid'}, 'finish_reason': 'length'}], 'usage': {'prompt_tokens': 37, 'completion_tokens': 4, 'total_tokens': 41}}
```

### DIFF
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -993,6 +993,26 @@ def format_saiga(
     return ChatFormatterResponse(prompt=_prompt.strip())
 
 
+# Chat format for Google's Gemma models, see more details and available models:
+# https://huggingface.co/collections/google/gemma-release-65d5efbccdbb8c4202ec078b
+@register_chat_format("gemma")
+def format_gemma(
+    messages: List[llama_types.ChatCompletionRequestMessage],
+    **kwargs: Any,
+) -> ChatFormatterResponse:
+    system_message = _get_system_message(messages)
+    if system_message is not None and system_message != "":
+        raise ValueError(
+            "`role='system'` messages are not allowed on Google's Gemma models."
+        )
+    _roles = dict(user="<start_of_turn>user\n", assistant="<start_of_turn>model\n")
+    _sep = "<end_of_turn>\n"
+    _messages = _map_roles(messages, _roles)
+    _messages.append((_roles["assistant"], None))
+    _prompt = _format_no_colon_single(system_message="", messages=_messages, sep=_sep)
+    return ChatFormatterResponse(prompt=_prompt, stop=_sep)
+
+
 # Tricky chat formats that require custom chat handlers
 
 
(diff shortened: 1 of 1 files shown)


alvarobartt:
P.S. I'm still unsure whether the `format_gemma` should raise a `ValueError` when the `system_prompt` is provided, as maybe instead if safer to just print out a warning stating that the `system_prompt` cannot be used i.e. will be ignored, WDYT @abetlen?

felipelo:
> P.S. I'm still unsure whether the `format_gemma` should raise a `ValueError` when the `system_prompt` is provided, as maybe instead if safer to just print out a warning stating that the `system_prompt` cannot be used i.e. will be ignored, WDYT @abetlen?

Agree ^^^

There are cases where a Prompt System is being used and happens to hold a 'system' value. I would even suggest to append the system_prompt on the first user message.

abetlen:
@alvarobartt thank you for the contribution.

As for system message I'm a little conflicted on this, the issue is that there are at least 3 entirely valid ways to handle it: raise an error, concat it to the user message, or ignore it entirely. For now I think ignoring a system message is the safest option (potentially raising a warning in verbose mode). In the future I'd like to migrate all of the chat templates to jinja2 then people can just pass custom templates easily to override these model defined presets.

alvarobartt:
> As for system message I'm a little conflicted on this, the issue is that there are at least 3 entirely valid ways to handle it: raise an error, concat it to the user message, or ignore it entirely. For now I think ignoring a system message is the safest option (potentially raising a warning in verbose mode). In the future I'd like to migrate all of the chat templates to jinja2 then people can just pass custom templates easily to override these model defined presets.

Fair, then I'll do that for the moment, and if you happen to open a draft PR with the Jinja2 migration I'm happy to help out! 🤝🏻

PR #1230, opened 2024-02-27
Fix eos/bos_token for Jinja2ChatFormatter

### LINKED ISSUE #1228: Automatically setting eos/bos_token from metadata does not work when falling back to Jinja2ChatFormatter
When falling back to Jinja2ChatFormatter in Llama.__init__() the following code erroneously sets eos_token and bos_token to blank strings, causing the Jinja2 template to malfunction and the model to not respond:

```python
                eos_token = self.detokenize([eos_token_id]).decode("utf-8")
                bos_token = self.detokenize([bos_token_id]).decode("utf-8")
```

If I change this piece of code to the following the Jinja2 template works as it should and the model responds correctly:

```python
                eos_token = self._model.token_get_text(eos_token_id)
                bos_token = self._model.token_get_text(bos_token_id)
```

However this might not be the appropriate solution, hence no PR.

CISC:
The token strings were not correctly retrieved (empty).

Closes #1228

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -408,8 +408,8 @@ class Llama:
                 except:
                     bos_token_id = self.token_bos()
 
-                eos_token = self.detokenize([eos_token_id]).decode("utf-8")
-                bos_token = self.detokenize([bos_token_id]).decode("utf-8")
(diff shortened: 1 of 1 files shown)


abetlen:
Thanks for the contribution!

Issue #1178, opened 2024-02-12
Improve installation process


abetlen:
Open to suggestions / assistance on how to make installation easier and less error prone.

One thought is to add better platform detection to the cmakelists and provide better docs / links if required environment variables aren't set / libraries can't be found.

ElliottDyson:
GPU detection to decide on what llama.cpp build to use? For example, if Arc is detected, but no Nvidia nor AMD card, then use the build that actively supports it.

vriesdemichael:
Wouldn't llama.cpp be the logical place to implement these improvements?

One thing improvement that would logically fall in the responsibility of llama-cpp-python is using the prebuilt llama.cpp when using windows, but other than that it would just be another layer of cmake

kristaller486:
Prebuilt wheels with GPU support for all platforms (on Github or PyPI). According to my observations, installing llama-cpp-python with GPU support is the most popular problem when installing llama-cpp-python, prebuilt packages should fix it.

abetlen:
> Prebuilt wheels with GPU support for all platforms (on Github or PyPI). According to my observations, installing llama-cpp-python with GPU support is the most popular problem when installing llama-cpp-python, prebuilt packages should fix it.

I'm partial to this, PyPI is a little annoying because we would need different package names for each but if we did it using seperate indexes (similar to pytorch) this should work. Ideally this would be done via seperate index URLs for metal, CUDA, etc. Maybe can be done with a GitHub pages run on each release.

DvitryG:
will the installation without wheels be supported? I just tried to update the package after a long break and got an error: `Getting requirements to build wheel did not run successfully.`

This is the first time I've come across wheels, so I want to ask if performance will deteriorate from using pre-build libraries? I don't have the most powerful computer, so it's important for me to make the most of it.

abetlen:
@DvitryG yes source installation will always be the default (`pip install llama-cpp-python` etc) and it should offer the most control / performance.

waheedi:
Thank you for making the effort to create these bindings, it would have been a nightmare ;D 

One important improvement, specifically  if building from a cloned source repo, is to keep the newly built wheel in place, maybe somewhere like `dist/llama_cpp_python-0.2.76-cp310-cp310-linux_x86_64.whl`

Currently there is no wheel you can grab in your hand but rather its directly installed into packages

neviaumi:
I recently try install `llama-cpp-python` on my project in X86 macbook pro.

[...]

Issue #741, opened 2023-09-20
Publish all wheels to PyPI


simonw:
It looks like PyPI only has the source distribution for each release: https://pypi.org/project/llama-cpp-python/0.2.6/#files

<img width="830" alt="CleanShot 2023-09-20 at 12 25 12@2x" src="https://github.com/abetlen/llama-cpp-python/assets/9599/e1dedc7e-d1e5-4c1d-8175-ab18187b0f48">

But the GitHub release at https://github.com/abetlen/llama-cpp-python/releases/tag/v0.2.6 lists many more files than that:

<img width="1086" alt="CleanShot 2023-09-20 at 12 25 51@2x" src="https://github.com/abetlen/llama-cpp-python/assets/9599/089a5103-6fbf-42af-ab3f-a7ec400de997">

Would it be possible to push those wheels to PyPI as well?

I'd love to be able to `pip install llama-cpp-python` and get a compiled wheel for my platform.

abetlen:
Hey @simonw! Big fan of your datasette project.

I hear you and I would like to make the setup process a little easier and less error-prone.

Currently llama.cpp supports a number of optional accelerations including several BLAS libraries, CUDA versions, OpenCL, and Metal. In theory I could build a pre-built wheel that just includes a version of llama.cpp with no real accelerations enabled but I feel like this is counterintuitive to the goal of providing users with the fastest local inference for their hardware.

I'm open to suggestions though, and I'll try to think of some possible solutions.

simonw:
The two approaches I can trying of that might work are:
- publish separate wheels for each platform, with separate names
- publish one large wheel that bundles the different versions together and has code that can pick the "right" one

For that first option, one way that could work is to have a `llama-cpp-python` package which everyone installs but which doesn't actually work until you install one of the "backend" packages: `llama-cpp-python-cuda-12` or `llama-cpp-python-metal` or similar.

How large are the different binaries? If all of them could be bundled in a single wheel that was less than 50MB then that could be a neat solution, if you can write code that can detect which one to use.

You could even distribute that as `llama-cpp-python-bundle` and tell people to install that one if they aren't sure which version would work best for them.

it's a tricky problem though! I bet there are good options I've missed here.

abetlen:
Hey @simonw it took a while but this is finally possible through a self-hosted PEP503 repository on Github Pages (see https://github.com/abetlen/llama-cpp-python/pull/1247)

You should now be able to specify

```bash
[...]

Issue #243, opened 2023-05-19
Precompiled wheels with CuBLAS activated


ParisNeo:
I use your library in my UI:
https://github.com/nomic-ai/gpt4all-ui

But Somehow, some users report that when they run my software and it tries to install your library, they get errors as the pip is trying to recompile the wheel.

Is there a way to precompile this for many lpatforms so that they won't need to build as this requires them to have a build environment which is not easy for noobs.

Thanks for this wonderful backend

gjmulder:
As per [@jmtatsch's reply](https://github.com/abetlen/llama-cpp-python/issues/236#issuecomment-1553277255) to my idea of pushing pre-compiled Docker images to Docker hub, providing precompiled wheels is likely equally problematic due to:

1. `llama.ccp` interrogating the hardware it is being compiled on and then aggressively optimises its compiled code to perform for _that_ specific hardware (e.g. ARM64 or x86_64 (and then _within_ x86_64 it may (or may not) use fp16, AVX, AVX2 and/or AVX512 [Intel hardware acceleration extensions](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions)))
2. If you pre-compile against CUDA, the exact same CUDA version needs to be pre-installed, unless you built a static binary that included all the CUDA libraries. And even then you might get NVidia GPU driver compatibility issues which I suspect may be the problem in issue #234

So you'd potentially need to pre-compile for every possible combination of hardware and CUDA version. Assuming (conservatively) there's say 32 (ARM64 + Intel) hardware combinations and [11 possible major CUDA versions](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-toolkit-major-component-versions), that's 352 wheels. Then you need to somehow educate the user on how to choose the single best wheel for his environment. :facepalm:

You could provide a static wheel with all AVX* disabled, no CUDA CuBLAS support but instead jsut with OpenBLAS, but then you'd get people complaining how much faster `llama.cpp` is :man_shrugging:

ParisNeo:
Not an easy task I confess.

gjmulder:
> Not an easy task I confess.

@ParisNeo on Discord trying to decide what all the moving parts are:

[gpt4all-webui](https://discord.com/channels/1076964370942267462/1093699802149171312)

AlphaAtlas:
[...]

PR #1247, opened 2024-03-03
Binary wheels

abetlen:
Adds initial support for releasing binary wheels. Uses github-pages to publish one index per backend (CPU, Metal, CUDA, etc) to a static site.

Planned release tags for this PR (because I can test these):
- `cpu`
- `cu121`
- `cu122`
- `cu123`
- `metal`

# Usage (won't work until merge)

To install a cpu only image

```bash
pip install llama-cpp-python --extra-index-url https://abetlen.github.io/llama-cpp-python/whl/cpu
```

# TODO

- Port the release workflows from https://github.com/jllllll/llama-cpp-python-cuBLAS-wheels
- Reduce number of wheels produced per tag

Related: #1178, #243 and #741


oobabooga:
I have been building llama-cpp-python wheels for each new release in my fork of jllllll's repository:

https://github.com/oobabooga/llama-cpp-python-cuBLAS-wheels/actions

It has some minor changes to avoid hitting rate limit errors. It also builds for all Mac versions (I think jllllll removed Mac 11 due to an error that used to happen but doesn't anymore).

abetlen:
@oobabooga yes I saw that, really appreciate you guys carrying that! I'll probably just port over those release workflows and modify them so they name the release based on the version and backend tag ie `v0.2.55-cu121` or something like that.

It would be great if we can reduce the number of builds per release though.

ElliottDyson:
Any chance intel arc binaries could also be added? Thanks

ElliottDyson:
> Any chance intel arc binaries could also be added? Thanks

With Intel's new IPEX-LLM release this is no longer necessary. A pointer to this however might be useful as it isn't well advertised.

Issue #1256, opened 2024-03-05
add flag to disable EventSourceReponse `ping` messages


khimaros:
**Is your feature request related to a problem? Please describe.**
some OpenAI based software such as https://github.com/ztjhz/BetterChatGPT/issues/538 do not gracefully handle `ping` messages. this prevents them from working reliably with `llama-cpp-python[server]`

**Describe the solution you'd like**
add a flag to `llama-cpp-python[server]` to disable ping messages, for use with apps which are not familiar with `:` prefixed comments.

**Describe alternatives you've considered**
could disable all ping messages in llama-cpp-python[server] but this may break some other use cases.

PR #1257, opened 2024-03-05
add `disable_ping_events` flag

khimaros:
for backward compatibility, this is false by default. it can be set to true to disable EventSource pings which are not supported by some OpenAI clients. fixes https://github.com/abetlen/llama-cpp-python/issues/1256


abetlen:
Hey @khimaros thank you for the contribution!

PR #1263, opened 2024-03-08
Switch embed to llama_get_embeddings_seq

iamlemec:
Due to updates in https://github.com/ggerganov/llama.cpp/pull/5796, sequence level embeddings are now output through a separate channel from token level embeddings, and they are accessed with `llama_get_embeddings_seq`.

### DIFF
--- a/llama_cpp/llama_cpp.py
+++ b/llama_cpp/llama_cpp.py
@@ -1803,22 +1803,6 @@ def llama_get_embeddings_ith(
     ...
 
 
-# // Get the embeddings for sequence seq_id
-# // shape: [n_embd] (1-dimensional)
-# LLAMA_API float * llama_get_embeddings_seq(struct llama_context * ctx, llama_seq_id seq_id);
-@ctypes_function(
-    "llama_get_embeddings_seq",
-    [llama_context_p_ctypes, ctypes.c_int32],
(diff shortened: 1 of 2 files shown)


abetlen:
@iamlemec thank you! Just so I understand, the sequence level embeddings are the ones that are pooled up to the end of the from the last processed batch?

Also, I think the new function in `llama_cpp.py` is duplicated by accident.

llama_cpp/llama_cpp.py
```diff
@@ -1803,6 +1803,22 @@ def llama_get_embeddings_ith(
...
+    ctypes.POINTER(ctypes.c_float),
+)
+def llama_get_embeddings_seq(
+    ctx: llama_context_p, i: Union[ctypes.c_int32, int], /
+) -> CtypesArray[ctypes.c_float]:
+    """Get the embeddings for sequence seq_id
+    shape: [n_embd] (1-dimensional)"""
+    ...
+
+
```
abetlen, line 1821:
```suggestion
```

I think this is a duplicate

iamlemec:
Oh yeah, I didn't see you added it already! Yup, it's the pooled embeddings by sequence for the last batch. So for both mean pooling and cls (first token) it works, and for non-pooling it returns null.

PR #1264, opened 2024-03-08
Check for existence of clip model path

kejcao:
The following code,

```py
from llama_cpp.llama_chat_format import Llava15ChatHandler
Llava15ChatHandler(clip_model_path='this-path-does-not-exist')
```

When ran aborts and yields a core dump.

```
$ python3 main.py
Aborted (core dumped)
```

This PR installs a more graceful exception with an informative error message.


abetlen:
@kejcao thank you for this!

PR #1273, opened 2024-03-14
Add tools/functions variables to Jinja2ChatFormatter

abetlen:
- Adds new variables `tools`, `tool_choice`, `function_call` and `functions` to jinja template.
- Adds support for function / tool calling response to all chat formats when making explicit function calls (not auto)

### DIFF
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -296,6 +296,183 @@ def _convert_completion_to_chat(
         return _convert_text_completion_to_chat(completion)
 
 
+def _convert_completion_to_chat_function(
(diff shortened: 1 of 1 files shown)


CISC:
Nice addition with the grammar.

CISC:
Would it make sense to add an option for a response parser that would populate tool_calls if one or more function calls are detected? That would perhaps make it simpler to have generic function calling without a lot of back and forth from the outside.

abetlen:
@CISC for now I'll just keep the scope of this PR to single explicit function / tool calling which is now supported for every chat format. This feature works similarly to json schema mode however it returns a properly formatted chat function message.

Issue #1285, opened 2024-03-18
Encounter this fatal error: 'ggml-common.h' file not found in M1 mac


yanqianglu:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [x] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [x] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

Please provide a detailed written description of what you were trying to do, and what you expected `llama-cpp-python` to do.

With Metal support, it should load a llm model properly.

# Current Behavior

Got this error when loading a llm model

```
ggml_metal_init: error: Error Domain=MTLLibraryErrorDomain Code=3 "program_source:3:10: fatal error: 'ggml-common.h' file not found
#include "ggml-common.h"
         ^~~~~~~~~~~~~~~
" UserInfo={NSLocalizedDescription=program_source:3:10: fatal error: 'ggml-common.h' file not found
#include "ggml-common.h"
         ^~~~~~~~~~~~~~~
}
llama_new_context_with_model: failed to initialize Metal backend
```

This seems to be an issue with llama.cpp but it should be fixed by https://github.com/ggerganov/llama.cpp/issues/5977. I'm using the v0.2.57 version of llama-cpp-python which should include the fix as well. However, I'm still seeing this issue. 

Here is how I re-installed llama-cpp-python (however, I'm not using conda, but the pyenv environment is new)

```
pip uninstall llama-cpp-python -y
CMAKE_ARGS="-DLLAMA_METAL=on" pip install -U llama-cpp-python --no-cache-dir
```

# Environment and Context

Please provide detailed information about your computer setup. This is important in case the issue is not reproducible except for under certain specific conditions.

* Physical (or virtual) hardware you are using, e.g. for Linux:

`$ lscpu`

* Operating System, e.g. for Linux:

`$ uname -a`

```
[...]

PR #1289, opened 2024-03-19
set LLAMA_METAL_EMBED_LIBRARY=on on MacOS arm64

bretello:
Fixes https://github.com/abetlen/llama-cpp-python/issues/1285

See https://github.com/ggerganov/llama.cpp/commit/381da2d9f0940d7009e3e918bed36338c8ff2fbb


abetlen:
@bretello thank you so much for this fix!

PR #1296, opened 2024-03-21
Make saved state more compact on-disk

tc-wolf:
Was looking at making the llama context more compact so that could store a cache for a large # of documents.

## Changes

- Only store up to `n_tokens` logits instead of full `(n_ctx, n_vocab)` sized array.
  - Difference between ~350MB and ~1500MB for example prompt with ~300 tokens vs. context size of 4096.
- Auto-formatting changes
  - Can back these out if not desired

Q: Do we need to store `Llama.scores` when logits are already saved on the `llama_context` struct?  This ~doubles the size of `LlamaState`, but I tried reloading directly using `get_logits` and the two were not the same (not sure why, think it could be repetition penalty applied to `Llama.scores`?  There wasn't a logits_bias or `logits_processor` being applied AFAICT).

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -18,6 +18,7 @@ from typing import (
     Iterator,
     Deque,
     Callable,
+    Dict,
 )
 from collections import deque
 from pathlib import Path
@@ -78,7 +79,9 @@ class Llama:
         n_batch: int = 512,
         n_threads: Optional[int] = None,
         n_threads_batch: Optional[int] = None,
-        rope_scaling_type: Optional[int] = llama_cpp.LLAMA_ROPE_SCALING_TYPE_UNSPECIFIED,
+        rope_scaling_type: Optional[
+            int
+        ] = llama_cpp.LLAMA_ROPE_SCALING_TYPE_UNSPECIFIED,
         pooling_type: int = llama_cpp.LLAMA_POOLING_TYPE_UNSPECIFIED,
         rope_freq_base: float = 0.0,
         rope_freq_scale: float = 0.0,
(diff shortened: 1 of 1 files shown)


abetlen:
Hey @tc-wolf thanks for the contribution. First do you mind reverting the formatting changes, thank you.

Next, yes we should only store the `(n_tokens, n_vocab)` sized logits array, however afaik it's untrue that this is stored in the `llama_context`. The `llama_context` only stores the logits for the last batch that was processed.

tc-wolf:
I'll work on putting together a better reprex that I can share, but I was able to reload and get logits for predicted tokens to match:

```python
def reload_from_low_level_state_file(model: Llama, state_file: str):
    from llama_cpp import llama_load_session_file

    n_ctx = model.n_ctx()
    session_tokens = (llama_cpp.llama_token * n_ctx)()
    n_tok = llama_cpp.ctypes.c_size_t()

    # Load context from file
    status = llama_load_session_file(
        model.ctx,
        state_file.encode("utf-8"),
        session_tokens,
        n_ctx,
        llama_cpp.ctypes.byref(n_tok),
    )

    if status != 1:
        raise ValueError(f"Failed to load context from '{state_file}'")

    n_tok_int = int(n_tok.value)
    print(f"n_tok: {n_tok_int}")

    assert (
        session_tokens[:n_tok_int] == state.input_ids[:n_tok_int].tolist()
    ), "Tokens should match"

    # Set various objects like in `load_state`
    n_vocab = model.n_vocab()
    logits_from_low_level = np.ctypeslib.as_array(
        model._ctx.get_logits(), (n_tok_int, n_vocab)
    )

    model.scores[:n_tok_int, :] = logits_from_low_level.copy()
    model.scores[n_tok_int:, :] = 0.0

    for i in range(0, n_tok_int):
        model.input_ids[i] = session_tokens[i]

    model.n_tokens = n_tok_int
```

and then predict with the same prompt, and got the same logits for predicted tokens:

```python
# Loads and resets to state created by doing `save_state` after `model.eval(prompt_tokens)`
model = get_nekomata7b()
model.load_state(state)

reload_from_low_level_state_file(model, "nekomata-7b-ctx.bin") 
out3 = model.create_chat_completion(
    [
        {"role": "system", "content": DEFAULT_SYSTEM_PROMPT},
        {"role": "user", "content": prompt},
    ],
    temperature=0.0,
    max_tokens=300,
    seed=1432
)
out3
pred_logits3 = model.scores[n_tok_prompt:model.n_tokens, :] 
assert (pred_logits == pred_logits3).all(), "All prediction logits should match"
```

This is without any logit bias and with no `logit_processor`s.

Issue #1305, opened 2024-03-26
cache-type-k and cache-type-p parameters support


riverzhou:
Origin llama.cpp is support cache-type-k and cache-type-p setting, but llama-cpp-python server is not.

Limour-dev:
[Setting quantization below f16](https://github.com/Limour-dev/llama-python-streamingllm/blob/main/llama_cpp_python_streamingllm.py#L383) will result in an error: GGML_ASSERT: llama.cpp\ggml.c:7553: false

abetlen:
@riverzhou thanks to @Limour-dev this should be in the latest v0.2.58 release by setting `llama_cpp.Llama(..., type_k=llama_cpp.GGML_TYPE_I8)`

PR #1307, opened 2024-03-28
add KV cache quantization options

Limour-dev:
https://github.com/abetlen/llama-cpp-python/discussions/1220
https://github.com/abetlen/llama-cpp-python/issues/1305


ddh0:
Would love for this to get merged!

abetlen:
Hey @Limour-dev thanks for the contribution.

A couple changes I made before merging:
- Added `ggml_type` int enum values to `llama_cpp.py` and used these instead of the mapped string names for kv quantization types.
- Added support for the `type_k` and `type_v` to the server as well.

PR #1317, opened 2024-03-31
Changed local API doc references to hosted

lawfordp2017:
Having API documentation references to a local server which may not be available yet (to those who are just evaluating) can be confusing.


abetlen:
Woops  @lawfordp2017 that was definitely a typo. Thank you for the fix!

PR #1333, opened 2024-04-08
add min_tokens argument

twaka:
Hi,
I'm added min_tokens argument which makes eos token's probability to -inf by logits processor when number of generated tokens is smaller than it.
Note that this implementation doesn't prevent stopping generation by another condition (e.g. max_tokens, stop, stopping_criteria).
Completes #240


twaka:
Thank you for taking a look into!
One problem I could think of is that MinTokensLogitProcessor needs prompt tokens' length.
It would not straightforward to obtain before tokenization occurs in llama_cpp._create_completion especially for chat completion.

abetlen:
Hey @twaka sorry for the delay, and thank you for implementing the change, happy to merge this now!

PR #1347, opened 2024-04-17
Improve function calling (auto selection, parallel functions, parameters grammar)

themrzmaster:
- update json to grammar based on https://github.com/ggerganov/llama.cpp/blob/master/examples/json_schema_to_grammar.py (now the required parameter works and not every function parameter is sampled)
- created a generic function calling template, which i created a vicuna derivative
- improved/fixed auto tool selection

Examples:
  ````
response = client.chat.completions.create(
    model="WizardLM-2-7B.Q8_0.gguf",
   messages = [
          {
            "role": "system",
            "content": "Always follow user requests"
          },
          {
            "role": "user",
            "content": "search for items with the term burguer and a maximum price of 10 dollars."
          },
         

        ],
      tools=[
           {
        "type": "function",
        "function": {
            "name": "search_merchant",
            "description": "Search for merchants in the catalog based on the term",
            "parameters": {
                "type": "object",
                "properties": {
                    "term": {
                        "type": "string",
                        "description": "Term to be searched for finding merchants.",
                    }
                },
                "required": ["term"],
            },
        },
    },
    {
        "type": "function",
        "function": {
            "name": "search_item",
            "description": "Search for items in the catalog based on various criteria.",
            "parameters": {
                "type": "object",
                "properties": {
                    "term": {
                        "type": "string",
                        "description": "Term to be searched for finding items, with removed accents.",
                    },
                    "item_price_to": {
                        "type": "integer",
                        "description": "Maximum price the user is willing to pay for an item, if specified.",
                    },
                    "merchant_delivery_fee_to": {
                        "type": "integer",
                        "description": "Maximum delivery fee the user is willing to pay, if specified.",
                    },
                    "merchant_payment_types": {
                        "type": "string",
                        "description": "Type of payment the user prefers, if specified.",
                        "enum": [
                            "Credit Card",
                            "Debit Card",
                            "Other",
                        ],
                    },
                },
                "required": ["term"],
            },
        },
    }],
       tool_choice="auto"
)
````

gives me:
[...]

PR #1351, opened 2024-04-17
Improve function calling (auto selection, parallel functions)

themrzmaster:
Following:https://github.com/abetlen/llama-cpp-python/pull/1347


abetlen:
@themrzmaster thank you, will review over the weekend. Additionally I'd like to add support for the Hermes 2 Pro format so I may add a few commits over this.

themrzmaster:
cool! I added llama 3 initial support. I am having trouble making it work with 70B, 8B is fine.

PR #1371, opened 2024-04-21
Add Llama-3 chat format

andreabak:
This PR adds support for the recently-released Llama-3 models by Meta. Specifically this chat format is used for their [`Instruct` pretrained models](https://huggingface.co/meta-llama/Meta-Llama-3-70B-Instruct).

Model card: https://github.com/meta-llama/llama3/blob/main/MODEL_CARD.md
See reference implementation for the chat format: https://github.com/meta-llama/llama3/blob/main/llama/tokenizer.py#L202-L229

### DIFF
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -35,6 +35,9 @@ MISTRAL_INSTRUCT_EOS_TOKEN = "</s>"
 # Source: https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1/blob/main/tokenizer_config.json
 MIXTRAL_INSTRUCT_CHAT_TEMPLATE = "{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if message['role'] == 'user' %}{{ '[INST] ' + message['content'] + ' [/INST]' }}{% elif message['role'] == 'assistant' %}{{ message['content'] + eos_token}}{% else %}{{ raise_exception('Only user and assistant roles are supported!') }}{% endif %}{% endfor %}"
 
+# Source: https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/blob/main/tokenizer_config.json
+LLAMA3_INSTRUCT_CHAT_TEMPLATE = "{% set loop_messages = messages %}{% for message in loop_messages %}{% set content = '<|start_header_id|>' + message['role'] + '<|end_header_id|>\n\n'+ message['content'] | trim + '<|eot_id|>' %}{% if loop.index0 == 0 %}{% set content = bos_token + content %}{% endif %}{{ content }}{% endfor %}{% if add_generation_prompt %}{{ '<|start_header_id|>assistant<|end_header_id|>\n\n' }}{% endif %}"
+
 ### Chat Completion Handler ###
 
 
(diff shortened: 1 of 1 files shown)


andreabak:
I made the system message be just like any other role, since from the reference code there doesn't seem to be a distinction between those.
Conventionally the first message is the system one, and usually no extra system ones are allowed within the dialog lines, but there doesn't seem to be such restriction in the llama-3 format (at least just from looking at the code).

masterofaimoein:
Please add Llama-3 chat format as soon as possible

masterofaimoein:
> This PR adds support for the recently-released Llama-3 models by Meta. Specifically this chat format is used for their [`Instruct` pretrained models](https://huggingface.co/meta-llama/Meta-Llama-3-70B-Instruct).
> 
> Model card: https://github.com/meta-llama/llama3/blob/main/MODEL_CARD.md
> See reference implementation for the chat format: https://github.com/meta-llama/llama3/blob/main/llama/tokenizer.py#L202-L229

Sir, please add chatformat llama-3 , llama-cpp-not work correctly with llama-3

andreabak:
@abetlen also added chat template for format auto-detect and bumped llama.cpp to latest version to properly support the eot token.
Rebased up to date with main.

N.B. unrelated issue, I noticed the LLAMA_CUBLAS cmake arg has been deprecated in favor of LLAMA_CUDA, so that should get changed in this codebase too eventually.

llama_cpp/llama_chat_format.py
```diff
@@ -920,6 +926,26 @@ def format_llama2(
...
+        system="<|start_header_id|>system<|end_header_id|>\n\n",
+        user="<|start_header_id|>user<|end_header_id|>\n\n",
+        assistant="<|start_header_id|>assistant<|end_header_id|>\n\n",
+    )
+    _begin_token = "<|begin_of_text|>"
+    _sep = "<|eot_id|>"
+    _messages = _map_roles(messages, _roles)
+    _messages.append((_roles["assistant"], None))
+    _prompt = _format_no_colon_single(_begin_token, _messages, _sep)
+    return ChatFormatterResponse(prompt=_prompt, stop=_sep)
```
jakekarnes42, line 946: We could consider adding `"<|im_end|>"` and `"<end_of_turn>"` as an additional stop tokens. I don't know if it's completely necessary, but ChatFormatterResponse looks like it accepts a list of stop tokens, and the llama.cpp project uses all three as stop tokens for Llama 3: https://github.com/ggerganov/llama.cpp/commit/8960fe86ae075c846c5df8848230d1904ba8877f

jakekarnes42 (changes requested):
Just some suggestions based on similar changes I saw in the [llama.cpp project](https://github.com/ggerganov/llama.cpp).

llama_cpp/llama_chat_format.py
```diff
@@ -35,6 +35,9 @@
 # Source: https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1/blob/main/tokenizer_config.json
 MIXTRAL_INSTRUCT_CHAT_TEMPLATE = "{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if message['role'] == 'user' %}{{ '[INST] ' + message['content'] + ' [/INST]' }}{% elif message['role'] == 'assistant' %}{{ message['content'] + eos_token}}{% else %}{{ raise_exception('Only user and assistant roles are supported!') }}{% endif %}{% endfor %}"
 
+# Source # TODO
```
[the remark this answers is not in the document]
andreabak: I added the TODO and forgot about it right after :sweat_smile: I'll add the link you suggested, thank you.

About the template I used the one that comes through from the gguf file and it matches and auto-detects the format successfully.

llama_cpp/llama_chat_format.py
andreabak: I think it makes sense to follow what's done in the llama.cpp library. One thing I'm not sure is if that code from the linked commit is only for llama-3 or more generic for any llama models? Because if it's the latter, maybe it's better to just stick to the `<|eot_id|>` that's explicitly defined in the released llama-3 code?

ramipellumbi:
Thanks for this PR! I have been using these changes with success locally to test Llama 3.

Issue #1288, opened 2024-03-19
(llama-cpp-python v0.2.57) RuntimeError: Failed to get embeddings from sequence pooling type is not set


Fuehnix:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [Y] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [Y] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [Y] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [Y] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

Please provide a detailed written description of what you were trying to do, and what you expected `llama-cpp-python` to do.
I tried running a simple hello world embedding query to make sure llama-cpp-python was working after doing a clean install of CentOS 9 with CUDA, Python 3.11.8, VSCode and the works.
Code:
```
 import argparse

from llama_cpp import Llama


llm = Llama(model_path="/home/jfuehne/Desktop/AI/Code/models/llama-2-13b-chat.Q5_K_M.gguf", embedding=True)

print(llm.create_embedding("Hello world!"))
```
Expected output (this is the output given for this code when I downgrade the llama-cpp-python package to 0.2.55):
```
ggml_init_cublas: GGML_CUDA_FORCE_MMQ:   no
ggml_init_cublas: CUDA_USE_TENSOR_CORES: yes
ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA RTX A6000, compute capability 8.6, VMM: yes
llama_model_loader: loaded meta data with 19 key-value pairs and 363 tensors from [../llama-2-13b-chat.Q5_K_M.gguf](https://file+.vscode-resource.vscode-cdn.net/../models/llama-2-13b-chat.Q5_K_M.gguf) (version GGUF V2)
llama_model_loader: Dumping metadata keys/values. Note: KV overrides do not apply in this output.
llama_model_loader: - kv   0:                       general.architecture str              = llama
llama_model_loader: - kv   1:                               general.name str              = LLaMA v2
llama_model_loader: - kv   2:                       llama.context_length u32              = 4096
[...]

PR #1380, opened 2024-04-24
Allow for possibly non-pooled embeddings

iamlemec:
Getting embeddings direct from "non-embedding" LLMs is pretty important for achieving SOTA performance (see [MTEB Leaderboard](https://huggingface.co/spaces/mteb/leaderboard)). This PR makes it so that the embed functions return token level embeddings when the `pooling_type` ends up being `LLAMA_POOLING_TYPE_NONE` and the appropriate sequence level embeddings otherwise. After that, the user can do whatever type of pooling they wish. This should address most of the concerns raised in #1288 too.

There are still some gotchas. If you request mean/cls pooling on a model that doesn't support it, it will hard crash in `llama.cpp`. And it's hard to know ex ante what types of pooling a model supports. So it's probably just better to wait until this gets sorted out in `llama.cpp`.

Also, this changes the default for `normalize` to `False`. You typically want to normalize after pooling, so you would not want to normalize when getting token level embeddings. The other option is defaulting to `False` for token level and `True` for sequence level, but that seems overly complicated.

*Note*: this depends on the `llama_pooling_type` function, which just got merged in https://github.com/ggerganov/llama.cpp/commit/b4e4b8a9351d918a56831c73cf9f25c1837b80d1.

### DIFF
--- a/llama_cpp/llama_cpp.py
+++ b/llama_cpp/llama_cpp.py
@@ -1179,6 +1179,12 @@ def llama_rope_type(model: llama_model_p, /) -> int:
     ...
 
 
+# LLAMA_API enum llama_pooling_type llama_pooling_type(const struct llama_model * model);
+@ctypes_function("llama_pooling_type", [llama_model_p_ctypes], ctypes.c_int)
+def llama_pooling_type(model: llama_model_p, /) -> int:
+    ...
+
+
 # LLAMA_API int32_t llama_n_vocab    (const struct llama_model * model);
 @ctypes_function("llama_n_vocab", [llama_model_p_ctypes], ctypes.c_int32)
 def llama_n_vocab(model: llama_model_p, /) -> int:
--- a/README.md
+++ b/README.md
@@ -575,7 +575,7 @@ llama = Llama(
 
 ### Embeddings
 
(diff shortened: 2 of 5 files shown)


llama_cpp/llama_cpp.py
```diff
@@ -1185,6 +1185,12 @@ def llama_rope_type(model: llama_model_p, /) -> int:
     ...
 
 
+# LLAMA_API enum llama_pooling_type llama_pooling_type(const struct llama_model * model);
+@ctypes_function("llama_pooling_type", [llama_model_p_ctypes], ctypes.c_int)
+def llama_pooling_type(model: llama_model_p, /) -> int:
+    ...
+
+
```
abetlen, line 1193:
```suggestion
```

I added this function elsewhere in the file to match the llama.h order so we can delete this.

abetlen (changes requested):

abetlen:
@iamlemec thank you so much for this!

I suppose expanding the possible return type of `Embeddings` is necessary to support sequences of embeddings and provide the ability to do your own pooling so I think that's okay.

Would you mind expanding the Embedding section in the README to explain this behaviour? It's quite bare at the moment and would be very helpful if anyone has questions here.

abetlen:
@iamlemec thank you that's fantastic! I'll go ahead and merge this now!

PR #1382, opened 2024-04-25
Add the Command R chat format

euxoa:


CISC:
This should not strictly be necessary as recent GGUFs have the chat format embedded (which will be automatically applied through Jinja2ChatFormatter), I've submitted a request in older repos on HF to be updated (and many of them have already done so).

If you have an outdated GGUF and don't wish to redownload it you can update your local file using the `gguf-new-metadata.py` script in `llama.cpp/gguf-py/scripts` and the latest Command R `tokenizer_config.json` from HF:
```bash
python gguf-new-metadata.py input.gguf output.gguf --chat-template-config tokenizer_config.json
```

uncodecomplexsystems:
@CISC 
There are some arguments however:
- As you said yourself, there is a lot (the vast majority to be honest) of GGUFs that don't have this yet
- lama-cpp-python already offers a lot of chat formats. llama-cpp also introduced the command-r chat format. As Command-R (Plus) is currently the most capable open models (or tie with llama3) I think it makes a lot of reason to merge this.
- It's just a minor merge to an existing function.
- Would really help a lot of people.

As soon as more GGUFs have the formats embedded the situation changes. But right I now this merge would just be super helpful. The model is a powerhouse for the open weights community.

Merge would be <3 <3 <3

CISC:
@uncodecomplexsystems As you say, it's just a minor merge, I'm not opposed to it, I'm just saying it's not strictly necessary. :)

euxoa:
> If you have an outdated GGUF and don't wish to redownload it you can update your local file [...]

Thanks, I didn't know that!

I have various GGUFs for Qwen-1.5, Command R, and Llama 3's, and the automatic setup of the chat format fails surprisingly often (including a 5-bit Llama-3 8B GGUF; Qwens become configured to `chatml` even though there is a `qwen` format  defined in the code). 

And confusingly, Command R kind of works with the `chatml` format and probably even with the default `llama-2` format, but then in tests suffers from poorer prompt following, and outputs <NAME> tags instead of names.

CISC:
> I thought those with `None` were fails, but do they actually get their chat format correctly from the template?

Yes, None means it found an embedded template (enable verbose and it will output the template), if no template can be guessed or found it will fall back to llama-2, see [llama.py](https://github.com/abetlen/llama-cpp-python/blob/266abfc1a36a4b33c0da3f3689676827df3e6a88/llama_cpp/llama.py#L436).

uncodecomplexsystems:
Based on the inactivity both in this PR and the phi3 one I suppose your stance @abetlen is to not merge any more new chat templates into llama-cpp-python, right?
I think it's important to know. Thx!

CISC:
@uncodecomplexsystems Patience, I'm sure there's just a lot going on (here or elsewhere) right now.

khimaros:
it's worth noting that llama.cpp/examples/server now has its own chat template handling https://github.com/ggerganov/llama.cpp/wiki/Templates-supported-by-llama_chat_apply_template, which i believe is based on the `llama_chat_template_apply()` API in llama.cpp. there is also a bit of flux there are a few PRs and issues seeking a more general solution:

https://github.com/ggerganov/llama.cpp/pull/6822
https://github.com/ggerganov/llama.cpp/pull/6834
https://github.com/ggerganov/llama.cpp/issues/4216
https://github.com/ggerganov/llama.cpp/issues/6726
https://github.com/ggerganov/llama.cpp/issues/5922

PR #1383, opened 2024-04-26
Add the Phi 3 mini chat format

reddiamond1234:
*my first PR :)


handshape:
It's getting weird over on HF. MS has revised the Phi-3 chat format to completely ignore the system role, and are telling consumers to rewrite their prompt structures instead.

https://huggingface.co/microsoft/Phi-3-mini-4k-instruct/discussions/51

0cc4m:
@abetlen Is there any way I can help? Would be nice to get the phi 3 format in, now that the larger models have been released.

llama_cpp/llama_chat_format.py
```diff
@@ -1325,6 +1331,27 @@ def format_gemma(
     _prompt = _format_no_colon_single(system_message="", messages=_messages, sep=_sep)
     return ChatFormatterResponse(prompt=_prompt, stop=_sep)
 
+@register_chat_format("phi_3_mini")
```
ihendley, line 1334: Better to call it just "phi-3" here and elsewhere

RomainMendez:
Any success in merging this ?

PR #1386, opened 2024-04-26
Proper fill-in-middle support

CISC:
Use prefix/middle/suffix tokens when metadata is present in GGUF, like f.ex. in [this](https://huggingface.co/CISCai/CodeQwen1.5-7B-Chat-SOTA-GGUF) one.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -940,18 +940,47 @@ class Llama:
 
         completion_id: str = f"cmpl-{str(uuid.uuid4())}"
         created: int = int(time.time())
+        prefix_token_id: Optional[int] = self.metadata.get("tokenizer.ggml.prefix_token_id")
+        middle_token_id: Optional[int] = self.metadata.get("tokenizer.ggml.middle_token_id")
+        suffix_token_id: Optional[int] = self.metadata.get("tokenizer.ggml.suffix_token_id")
         # If prompt is empty, initialize completion with BOS token to avoid
         # detokenization including a space at the beginning of the completion
         completion_tokens: List[int] = [] if len(prompt) > 0 else [self.token_bos()]
         # Add blank space to start of prompt to match OG llama tokenizer
         prompt_tokens: List[int] = (
             (
-                self.tokenize(prompt.encode("utf-8"), special=True)
-                if prompt != ""
-                else [self.token_bos()]
+                [int(prefix_token_id)]
+                if prefix_token_id and suffix is not None
+                else []
             )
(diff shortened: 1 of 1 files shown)


CISC:
See [this](https://github.com/ggerganov/llama.cpp/pull/6778) PR on how to add this metadata to GGUFs that do not have them (recent CodeLlama and CodeGemma GGUFs should have gotten them automatically on conversion).

llama_cpp/llama.py
```diff
@@ -940,18 +940,47 @@ def _create_completion(
 
         completion_id: str = f"cmpl-{str(uuid.uuid4())}"
         created: int = int(time.time())
+        prefix_token_id: Optional[int] = self.metadata.get("tokenizer.ggml.prefix_token_id")
+        middle_token_id: Optional[int] = self.metadata.get("tokenizer.ggml.middle_token_id")
+        suffix_token_id: Optional[int] = self.metadata.get("tokenizer.ggml.suffix_token_id")
         # If prompt is empty, initialize completion with BOS token to avoid
         # detokenization including a space at the beginning of the completion
         completion_tokens: List[int] = [] if len(prompt) > 0 else [self.token_bos()]
         # Add blank space to start of prompt to match OG llama tokenizer
```
CISC, line 949: What blank space does this comment refer to?

abetlen:
@CISC thank you for implementing this. And yeah lot's of moving pieces in this method (honestly very long overdue for a refactor) so avoiding doing anything fancy is ideal.

llama_cpp/llama.py
abetlen: It's to avoid a bug when generating from a completely empty prompt that causes a leading space to always be added but I think it may actually be incorrect for non-llama tokenizers, I'll have to check.

llama_cpp/llama.py
```diff
@@ -940,18 +940,47 @@ def _create_completion(
...
-                if prompt != ""
-                else [self.token_bos()]
+                [prefix_token_id]
+                if prefix_token_id >= 0 and suffix is not None
+                else []
+            )
+            +
+            (
+                (
+                    self.tokenize(prompt.encode("utf-8"), special=True)
```
CISC, line 959: This should probably be called with add_bos=False (see fddf3194e92d5e6419408b28f1307855d21c0c0b ) to avoid spurious special tokens. In fact not entirely sure if it makes sense to insert BOS (or any other special tokens), nor where to insert them, when using fill-in-middle?

PR #1391, opened 2024-04-27
Update dependabot.yml

Smartappli:
Add github-actions update


abetlen:
@Smartappli thank you!

Issue #1336, opened 2024-04-08
Models with multiple chat templates


CISC:
Not an issue yet, but will need to be handled once [this](https://github.com/ggerganov/llama.cpp/issues/6484) is implemented based on recent transformers [changes](https://github.com/huggingface/transformers/pull/29658).

Also note the kwargs change in the same PR which will be used by f.ex. C4AI Command R models (new chat template is not merged yet) to pass along `tools` and `documents`, while we already support `tools`, it might be worthwhile to support other things.

CISC:
Any suggestions on how to approach this? It has been merged in llama.cpp a while now, and many GGUFs already have the new metadata.

I suppose adding f.ex. a chat_template_name parameter and applying the chosen template (if found - should also output which templates are available (from tokenizer.chat_templates list) I guess) would be the initial step.

For server this gets more complicated, it would probably make sense to allow the caller to chose a template, and then also have an endpoint to see which templates are available?

Finally, how would you go about adding support for additional parameters to the template, like `documents` in the `rag` template?

abetlen:
@CISC do you mind posting a gguf that uses this right now.

Yeah I think we can do even more simple and not introduce any new parameters just use the existing `chat_format`. The default when no chat_format is specified is to use `default` then the others can just be specified there by name.

The chat formats will be accessible through the metadata not sure if we need to add anything new there but we should add an option to change chat format after initialization (I believe this has already been requested before).

CISC:
Sure, [pmysl](https://huggingface.co/pmysl/c4ai-command-r-plus-GGUF) was the first one to update their quants. If R+ is a bit too hefty, try [LlamaEdge](https://huggingface.co/second-state/C4AI-Command-R-v01-GGUF)'s Command R quant.

My main worry about using chat_format is that it might conflict with an existing choice, albeit unlikely.

abetlen:
@CISC good point, let's prefix these dynamically loaded chat templates with `chat_template` so `chat_template.rag` or `chat_template.tool_use` for the cohere model.

CISC:
@abetlen That seems reasonable, I'm thinking registering `chat_template.default` etc. as chat format at init with the Jinja2 handler setup done as fallback today and then just fall back to `chat_template.default`(if registered) instead.

CISC:
WIP changes worth paying attention to: huggingface/transformers#30621

CISC:
[...]

PR #1396, opened 2024-04-27
Support multiple chat templates - step 1

CISC:
As a first step towards #1336, allow user to to select template from metadata with chat_format parameter in the form of `chat_template.name`.


abetlen:
@CISC thank you I'll review this weekend!

abetlen:
@CISC sorry for the delay in reviewing, one thought I had is that I'm a little nervous about registering these templates globally on model load, this seems like a side-effect that will cause some headaches if users are unaware.

CISC:
Sure, I get that, it also just occurred to me that this will wreak havoc if a user loads more than one model with chat template metadata. :(

Maybe register it per-model somehow?

CISC:
@abetlen How about this? Does not clobber global registry, but still easily accessible for run-time switch.

abetlen:
@CISC perfect, I did update `chat_handlers` -> `_chat_handlers` to keep it private for now but otherwise it looks good to me.

Issue #1405, opened 2024-04-29
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'name' when running Functions.ipynb with Jupyter lab


southkorea2013:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [ ] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [ ] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [ ] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [ ] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

Output the temperature of San Francisco, Tokyo, and Paris

# Current Behavior
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'name'
INFO:     127.0.0.1:61901 - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error

# Environment and Context

Please provide detailed information about your computer setup. This is important in case the issue is not reproducible except for under certain specific conditions.

* Physical (or virtual) hardware you are using, e.g. for Linux:

`$ lscpu`

* Operating System, e.g. for Linux: macOS 14.3

`$ uname -a`

* SDK version, e.g. for Linux:

```
$ python3 --version
$ make --version
$ g++ --version
```

# Failure Information (for bugs)

jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'name'
INFO:     127.0.0.1:61901 - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error

# Steps to Reproduce

Please provide detailed steps for reproducing the issue. We are not sitting in front of your screen, so the more detail the better.

1. python3 -m llama_cpp.server --model /Users/***/models/functionary-7b-v2-GGUF/functionary-7b-v2.q4_0.gguf --chat_format functionary-v2 --hf_pretrained_model_name_or_path /Users/***/models/functionary-7b-v2-GGUF --n_gpu_layers 16
2. jupyter lab
3. execute the cell of Function Calling with OpenAI Python Client with Jupyter Lab
4. etc.

[...]

PR #1407, opened 2024-04-29
fix: add binding for name in ChatCompletionRequestToolMessage

JDScript:
Follow issue #1405

### DIFF
--- a/llama_cpp/llama_types.py
+++ b/llama_cpp/llama_types.py
@@ -222,6 +222,7 @@ class ChatCompletionRequestToolMessage(TypedDict):
     role: Literal["tool"]
     content: Optional[str]
     tool_call_id: str
+    name: Optional[str]
 
 
 class ChatCompletionRequestFunctionMessage(TypedDict):


abetlen:
@JDScript I'm not sure that's correct based on the OpenAI OpenAPI spec [here](https://github.com/openai/openai-openapi/blob/38c9c20c41871d316ec3a516c556db69f2900e1d/openapi.yaml#L7426)

JDScript:
Yeah, but I checked the document [here](https://platform.openai.com/docs/guides/function-calling), under Example invoking multiple function calls in one response. Examples given by OpenAI also adds the name parameter, which is the same as in the [examples/notebooks/Functions.ipynb](https://github.com/abetlen/llama-cpp-python/blob/main/examples/notebooks/Functions.ipynb). Therefore, I used an optional binding.

pseudorandomrants:
Any update on this? I am facing same issue while using functionary.

allenporter:
This does not seem like the correct fix.

I see this message when specifying a message `"role": "tool"`, so you can workaround with another role e.g. `system`

allenporter:
Also discussed here https://github.com/openai/openai-python/issues/1078

I see, the `tool` role template is expecting a name in the chat template:
- https://huggingface.co/meetkai/functionary-7b-v2/blob/68795699d4632529a92530ab55bbe8efbd62707a/tokenizer_config.json#L67
- https://huggingface.co/meetkai/functionary-small-v2.5/blob/89e6716271825c203b7e9c7836d2d3b1ed191bea/tokenizer_config.json#L2053

So a name needs to be populated from somewhere, or the `tool` role can't be used.

allenporter:
Perhaps we have to get it by looking at the history of tool calls with the same tool id.

PR #1419, opened 2024-05-01
Implement streaming for Functionary v2 + Bug fixes

jeffrey-fong:
**Features**
- Set up streaming for functionary-v2 chat handler
- Raises error if users use functionary-v1
- Returns only 1 function/tool call if `functions` is provided and >= 1 if `tools` is provided

**Bug Fixes**
- Fix `function-call={...}` always converted to `auto` bug


abetlen:
@jeffrey-fong thank you, looks good to me!

PR #1424, opened 2024-05-03
Fix: Propagate flash attn to model loader

dthuerck:
I noticed that even though setting `flash_attn` to `true` in my model config file, `llama.cpp` kept reporting `llama_new_context_with_model: flash_attn = 0`. This super-small PR fixes that - turns out the setting wasn't passed on to the model loader.


abetlen:
@dthuerck thank you!

PR #1427, opened 2024-05-03
LLaMA cpp python server: IPV6 support

Smartappli:
Replace uvicorn by hypercorn to support ipv6


abetlen:
Hey @Smartappli let me consider this, technically you should be able to do by just launching the app directly from the cli via hypercorn correct?

Smartappli:
Yes correct

Smartappli:
@gaby review plz

PR #1430, opened 2024-05-05
adding missing args in create_completion for functionary chat handler

skalade:
Looks like in all cases where `stream == False` the `create_completion` function was missing the new args.

I was getting this error when querying functionary v2 on llama-cpp-server:
```
TypeError: create_completion() missing 2 required positional arguments: 'prompt' and 'grammar'
```


jeffrey-fong:
Thanks for this! 🙏

PR #1439, opened 2024-05-09
Avoid duplicate special tokens in chat formats

### LINKED ISSUE #1501: Llama 3 Double BOS
# Prerequisites

- [x] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [x] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior
Run `create_chat_completion` on Llama 3 (in gguf format). 
```python
llm = Llama.from_pretrained(
    repo_id="bartowski/Meta-Llama-3-8B-Instruct-GGUF",
    filename="*Q6_K.gguf",
    verbose=True,
)
messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {
        "role": "user",
        "content": "Q: Name the planets in the solar system? A: Mercury, Venus, Earth, Mars, Jupiter, Saturn, Uranus, Neptune and Pluto."
    }
]
llm.create_chat_completion(
    messages=messages,
    max_tokens=2048,
)
```

# Current Behavior
Get the warning
```
llama_tokenize_internal: Added a BOS token to the prompt as specified by the model but the prompt also starts with a BOS token. So now the final prompt starts with 2 BOS tokens. Are you sure this is what you want?
```

# Environment and Context

Mac M1
Python 3.11.6

The warning is gone when applying the following chat template manually and using `create_completion` to get the response:
```python
template = "{% set loop_messages = messages %}{% for message in loop_messages %}{% set content = '<|start_header_id|>' + message['role'] + '<|end_header_id|>\n\n'+ message['content'] | trim + '<|eot_id|>' %}{% if loop.index0 == 0 %}{% set content = content %}{% endif %}{{ content }}{% endfor %}{% if add_generation_prompt %}{{ '<|start_header_id|>assistant<|end_header_id|>\n\n' }}{% endif %}"
```
The difference from the default Llama 3 template is that `set content = bos_token + content` is changed to `set content = content`.

Based on that, it seems the double BOS token is coming from the chat template applying the BOS token, but `create_completion` (probably when calling `tokenize`) is additionally adding the BOS token.

CISC:
Having multiple BOS can ruin generation, this would occur in several ways, usually through user adding them unnecessarily, in this case output a warning if we detect two in a row.

However, several formats/templates also have BOS in them, and we would add another one later in completion, therefore tokenize format/template prompts before completion to avoid this.

Added a new `added_special` property to `ChatFormatterResponse` to detect when BOS/EOS has already been added to prompt.

Also added missing `token_cls`, `token_sep`, ~~`add_bos_token` and `add_eos_token`~~ to `_LlamaModel` because I thought I needed them (turns out I didn't).

Fixes #1501

### DIFF
--- a/llama_cpp/_internals.py
+++ b/llama_cpp/_internals.py
@@ -170,14 +170,6 @@ class _LlamaModel:
         assert self.model is not None
         return llama_cpp.llama_token_eot(self.model)
 
-    def add_bos_token(self) -> int:
-        assert self.model is not None
-        return llama_cpp.llama_add_bos_token(self.model)
-
-    def add_eos_token(self) -> int:
-        assert self.model is not None
-        return llama_cpp.llama_add_eos_token(self.model)
-
     # Tokenization
 
     def tokenize(self, text: bytes, add_bos: bool, special: bool):
--- a/llama_cpp/_internals.py
+++ b/llama_cpp/_internals.py
@@ -142,6 +142,14 @@ class _LlamaModel:
         assert self.model is not None
         return llama_cpp.llama_token_eos(self.model)
 
+    def token_cls(self) -> int:
+        assert self.model is not None
+        return llama_cpp.llama_token_cls(self.model)
+
+    def token_sep(self) -> int:
+        assert self.model is not None
+        return llama_cpp.llama_token_sep(self.model)
+
     def token_nl(self) -> int:
         assert self.model is not None
         return llama_cpp.llama_token_nl(self.model)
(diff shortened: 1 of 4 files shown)


abetlen:
@CISC thank you (also for the other PRs), this looks good. I may update the warning to use the python `warnings` library (would be useful for deprecations as well) so it can be silenced in a standard way.

BigCatGit:
me too

Rocketknight1:
Hi, Matt from Hugging Face here. Just a note that we expect chat templates to contain all the special tokens the model expects in, and therefore in Transformers, **whenever a message has been formatted with a chat template, it should be tokenized without adding additional special tokens**.

We really had no choice but to do it this way - many base models add EOS at the end of sequences, but this will always cause problems if we want to generate a chat completion following a user message, as user messages often don't end with EOS in many chat formats.

CISC:
@Rocketknight1 Thanks for checking in. :)

Yeah, I perfectly understand the reasoning, it just wasn't obvious until recently that this and some built in chat formats were causing duplicate tokens due to this.

PR #1440, opened 2024-05-09
Support multiple chat templates - step 2

CISC:
Next step towards #1336 adds a new parameter to be able to pass arbitrary arguments to the template, much like transformers, except through an explicit parameter instead of just plain kwargs.


CISC:
@abetlen PTAL, `chat_template_kwargs` is important now for f.ex. disabling thinking.

PR #1441, opened 2024-05-10
Render templates in immutable sandbox

CISC:
Chat templates are rendered with `ImmutableSandboxedEnvironment` in `transformers` so no need to do otherwise here.


CISC:
Heh, what a fluke with dual PR, just noticed the discrepancy when double checking rendering with `transformers`...

BTW, the security PR left loader=jinja2.BaseLoader() in, it doesn't really make a difference, but it's not necessary since we're always using `.from_string()` which will not invoke a loader.

PR #1449, opened 2024-05-12
updated llama.cpp to b2861

Man2Dev:
fixes cmake bug


abetlen:
@Man2Dev thank you for the contribution however I usually handle the submodule updates just so I can manually review the upstream changes.

As for the issue, I see there's a fix for the version check, was it failing to set the correct cmake min version requirement?

Man2Dev:
> @Man2Dev thank you for the contribution however I usually handle the submodule updates just so I can manually review the upstream changes.
> 
> As for the issue, I see there's a fix for the version check, was it failing to set the correct cmake min version requirement?

I didn't check i was trying to update the python-llama-cpp package in fedora and saw fix come in right after the submodule update so i thought i would point it out.

PR #1465, opened 2024-05-16
example: LLM inference with Ray Serve

rgerganov:
This could be used as a starting point for people who want to experiment with distributed LLM inference on Ray


Smartappli (approved):

abetlen:
Awesome, thanks @rgerganov !

PR #1476, opened 2024-05-23
Improve Llama.eval efficiency

thoughtp0lice:
In llama_cpp/llama.py, the eval function converts return value of `self._ctx.get_logits()`, which is a CtypesArray, to list then copy it into `self.scores`. Here the CtypesArray is directly converted to a numpy array which speeds up the conversion and copying. The speed-up is especially noticeable on smaller models with faster inference time.


abetlen:
@thoughtp0lice thank you, that's perfect!

PR #1486, opened 2024-05-26
fix: match Transformers `tojson` in chat template rendering

CISC:
Updates this PR to only carry the remaining chat-template compatibility gap against current main:

- Override Jinja's `tojson` filter to match Transformers behavior (`ensure_ascii=False` by default and no HTML escaping).
- Keep the existing generation-tag and loopcontrols extensions intact.
- Add the matching changelog entry.


CISC:
Since the merging of a similar patch to `transformers` and the formalizing of `tools` (see references in #1336) there are starting to pop up a lot more models using `tojson` in their chat templates.

CISC:
@abetlen Even more `transformers` changes incoming, see updated OP.

CISC:
@abetlen It would be really nice to have this merged soon, it's starting to cause issues with quite a few templates not to have this in place.

Additionally this is a requirement for implementing inverse templates (which I'd like to do, as soon as it's merged in transformers, since it will be super useful for universal function calling and parsing).

serhii-nakon:
Hello I support to merge it. I have issues with Mistral models (devstral too)

CISC:
@abetlen I see `strftime_now` support was finally added, unfortunately that is not enough, PTAL.

PR #1492, opened 2024-05-29
Support SPM infill

CISC:
Add `spm_infill` option to perform infill in the Suffix/Prefix/Middle pattern instead of Prefix/Suffix/Middle as some models (like the new `Codestral`) prefer this.

Also added a tokenizer hack to remove leading space in suffix to improve inference, tested on several different tokenizers/vocabs with success.

### DIFF
--- a/examples/high_level_api/high_level_api_infill.py
+++ b/examples/high_level_api/high_level_api_infill.py
@@ -1,0 +1,33 @@
+import argparse
+
+from llama_cpp import Llama
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-m", "--model", type=str, default="../models/7B/ggml-models.bin")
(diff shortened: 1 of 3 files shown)


CISC:
Actually, since I ended up not using them there let's make it simple and remove them from that PR and add them here.

CISC:
My [Codestral GGUFs](https://huggingface.co/CISCai/Codestral-22B-v0.1-SOTA-GGUF) are up for those who wish to test with this option.

NOTE: It requires ggerganov/llama.cpp#7644 to not insert garbage `middle` token!

abetlen:
@CISC looks good overall, do you mind adding something minimal to `/examples` that would help me review this and demonstrat the FIM API?

CISC:
Sure, no problem.

CISC:
@abetlen Example added.

PR #1503, opened 2024-06-02
Support parallel function calls with tool_choice

CISC:
Enable the return of parallel function calls when `tool_choice` is set to a specific function, previously you would only get the following response for `get_current_weather` and the prompt `What's the weather like in Oslo and Stockholm?`:
```json
[
    {
        "id": "call__0_get_current_weather_cmpl-...",
        "type": "function",
        "function": {
            "name": "get_current_weather",
            "arguments": "{\"location\": \"Oslo, NO\", \"unit\": \"celsius\"}"
        }
    }
]
```

Now you will get all the `get_current_weather` calls:
```json
[
    {
        "id": "call__0_get_current_weather_cmpl-...",
        "type": "function",
        "function": {
            "name": "get_current_weather",
            "arguments": "{\"location\": \"Oslo, NO\", \"unit\": \"celsius\"}"
        }
    },
    {
        "id": "call__1_get_current_weather_cmpl-...",
        "type": "function",
        "function": {
            "name": "get_current_weather",
            "arguments": "{\"location\": \"Stockholm, SE\", \"unit\": \"celsius\"}"
        }
    },
]
```

Additionally the function call names will now be verified (and filtered if not matching), whereas before there was the possibility of picking up a different function call with similar parameters.

### DIFF
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -330,10 +330,49 @@ def _convert_completion_to_chat_function(
     ],
     stream: bool,
 ):
+    def _completion_text_to_tool_calls(
+        tool_name: str,
+        completion_text: str,
+        completion_id: str
+        stream: bool,
+    ) -> Union[
+        llama_types.ChatCompletionMessageToolCalls, List[llama_types.ChatCompletionMessageToolCallChunk]
+    ]:
+        try:
+            function_calls = json.loads(completion_text)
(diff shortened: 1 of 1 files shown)


lsorber:
I opened https://github.com/abetlen/llama-cpp-python/pull/1884 with a number of improvements to `chatml-function-calling`, including streaming tool use and fixing parallel function calling.

PR #1509, opened 2024-06-05
Integrate Functionary v2.5 + Refactor Functionary Code

jeffrey-fong:
- Implemented abstract base classes for FunctionaryChatHandler based on version for better maintenance.
- Changed chat_format to `functionary` instead of `functionary-v1`, `functionary-v2`, `functionary-v2.5`. I changed the old chat_format for the old handler to `functionary-old`. Hope it is not breaking any code at your side. Anyway, can we remove the old handler function already?
- Fully integrate Functionary v2.5 for both streaming and non-streaming.
- Added code_interpreter feature for v2 and v2.5. Auto activates when user passes in `{"type": "code_interpreter"}` in tools. Returns a tool call to a tool called `python` with the argument being the generated code.
- Functionary v1 also supports streaming now.

### DIFF
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
 import os
 import json
 import ctypes
@@ -1329,7 +1330,7 @@ def format_gemma(
 # Tricky chat formats that require custom chat handlers
 
 
-@register_chat_completion_handler("functionary")
+@register_chat_completion_handler("functionary-old")
 def functionary_chat_handler(
     llama: llama.Llama,
     messages: List[llama_types.ChatCompletionRequestMessage],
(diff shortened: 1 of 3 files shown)


allenporter:
I have tested this PR and it seems to be working just fine. 



I used this `models.json`:

```json
{
    "host": "0.0.0.0",
    "port": 8080,
    "models": [
      {
        "model": "functionary-small-v2.5.Q4_0.gguf",
        "model_alias": "functionary-small-v2.5-auto-tokenizer",
        "hf_model_repo_id": "meetkai/functionary-small-v2.5-GGUF",
        "hf_pretrained_model_name_or_path": "meetkai/functionary-small-v2.5-GGUF",
        "chat_format": "functionary",
        "n_gpu_layers": 32,
        "offload_kqv": true,
        "n_ctx": 8192,
        "use_mlock": false
      }
    ]
  }
```

Start server with `CONFIG_FILE=models.json  python3 -m llama_cpp.server` then calling with an openai client library, something like this:

```python
BASE_URL = "http://0.0.0.0:8080/v1"
API_KEY = "sk-xxx"
MODEL_ID = "functionary-small-v2.5-auto-tokenizer"

client = OpenAI(base_url=BASE_URL, api_key=API_KEY)

messages = [
    {"role": "user", "content": "what's the weather like in Hanoi?"}
]
tools = [ # For functionary-7b-v2 we use "tools"; for functionary-7b-v1.4 we use "functions" = [{"name": "get_current_weather", "description":..., "parameters": ....}]
    {
        "type": "function",
        "function": {
            "name": "get_current_weather",
            "description": "Get the current weather",
            "parameters": {
                "type": "object",
                "properties": {
                    "location": {
                        "type": "string",
                        "description": "The city and state, e.g., San Francisco, CA",
                    }
                },
                "required": ["location"]
            }
        }
    }
]

chat_response = client.chat.completions.create(
    model=model,
    messages=messages,
    tools=tools,
)
print(yaml.dump(chat_response.model_dump(), sort_keys=False, explicit_start=True))
```

Gives the following response:
```yaml
---
id: chatcmpl-0a698bb0-d38a-4a77-b2e6-2226e09f71c9
choices:
- finish_reason: tool_calls
  index: 0
  logprobs: null
  message:
    content: null
    role: assistant
    function_call: null
    tool_calls:
    - id: call_ePVfb4Gg0aNbCV41cKc9tkAi
      function:
        arguments: '{"location": "Hanoi"}'
        name: get_current_weather
      type: function
created: 1720904370
model: functionary-small-v2.5-auto-tokenizer
object: chat.completion
service_tier: null
system_fingerprint: null
usage:
  completion_tokens: 12
  prompt_tokens: 126
  total_tokens: 133
```

llama_cpp/llama_chat_format.py
```diff
@@ -1686,9 +1687,731 @@ def message_to_str(msg: llama_types.ChatCompletionRequestMessage):
...
+                        "index": 0,
+                        "finish_reason": finish_reason,
+                        "logprobs": logprobs,
+                        "delta": delta
+                    }
+                ],
+            )
+        
+        chunk_id, chunk_created = None, None
+        delta = {"role": None, "content": None, "function_call": None, "tool_calls": None}
```
allenporter, line 1872: Consider simplifying how `delta` is managed given  that it is mutable and a lot of care needs to be taken to make sure that all other fields are cleared. It might be simpler to fully create the delta each time, then fill in the missing fields inside of `yield_response` when creating the response. This also simplifies the final response.

llama_cpp/llama_chat_format.py
```diff
@@ -1686,9 +1687,731 @@ def message_to_str(msg: llama_types.ChatCompletionRequestMessage):
...
+                    {
+                        "index": 0,
+                        "finish_reason": finish_reason,
+                        "logprobs": logprobs,
+                        "delta": delta
+                    }
+                ],
+            )
+        
+        chunk_id, chunk_created = None, None
```
allenporter, line 1871: Given these are only used in the `If "auto"` case, push the scope of the variables inside that if statement.

llama_cpp/llama_chat_format.py
```diff
@@ -1686,9 +1687,731 @@ def message_to_str(msg: llama_types.ChatCompletionRequestMessage):
...
+        if not self.stream:
+            completion_or_completion_chunks["choices"][0]["text"] = completion_or_completion_chunks["choices"][0]["text"].lstrip()
+        
+        return _convert_completion_to_chat(completion_or_completion_chunks, stream=self.stream)  # type: ignore
+    
+    def generate(self, prompt: str, tool_func_choice: Optional[Union[str, Dict]]):
+        content = ""
+        calls = []
+        completion_tokens = 0
+        # If tool_choice/function_call is provided
```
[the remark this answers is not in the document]
allenporter: I think this can also simplify the streaming state management too.

allenporter (changes requested):
I'm not super familiar with the code base, but motivated to use this feature so thought I would help with review. Great work and thank you for contributing this as I am very eager to have a working function calling (I have trouble with earlier versions so hoping this version is more robust)

allenporter:
Also I realize some of my comments are on existing code which i realize the intent was not to change to make review easier --  but given how much is changing already it may be worth simplifying where possible.

llama_cpp/llama_chat_format.py
```diff
@@ -1686,9 +1687,731 @@ def message_to_str(msg: llama_types.ChatCompletionRequestMessage):
...
+            return prompt, completion_tokens, "content"
+        
+    def generate_content(self, prompt: str, content: str, completion_tokens):
+        stops = [self.tool_call_token, self.stop_token]
+        completion = self.create_completion(prompt=prompt, stop=stops, grammar=None)
+        completion_text = completion["choices"][0]["text"]
+        completion_tokens += completion["usage"]["completion_tokens"]
+        prompt += completion_text
+        content_prefix = "<|start_header_id|>assistant<|end_header_id|>\n\n"
+        content += prompt[prompt.rindex(content_prefix) + len(content_prefix):] + completion_text
```
allenporter, line 2333: When testing this out I see responses like `The smart curtains have been closed. smart curtains have been closed.`.

I am not sure, but from what I can see here this first appends the `completion_text` to the `prompt`, then gets the tail of the `prompt` and adds `completion_text` again to the `content`.  Does this mean `completion_text` was added twice?

llama_cpp/llama_chat_format.py
```diff
@@ -1686,9 +1687,731 @@ def message_to_str(msg: llama_types.ChatCompletionRequestMessage):
...
+        elif len(completion_text) == 0:
+            return delta_texts, prompt, "stop"
+        else:
+            return delta_texts, prompt, "content"
+        
+    def generate_streaming_content(self, prompt: str):
+        stops = [self.tool_call_token, self.stop_token]
+        completion = self.create_completion(prompt=prompt, stop=stops, grammar=None)
+        for chunk in completion:
+            delta_text = chunk["choices"][0]["text"]
```
allenporter, line 2385: Does `generate_streaming_content` for v2.5 also need to strip the `content_prefix` assistant message?

llama_cpp/llama_chat_format.py
allenporter: Applying this fix *seems* to address the issue:

```suggestion
        content += prompt[prompt.rindex(content_prefix) + len(content_prefix):]
```

axel7083 (commented):
When using the example from the [Functions.ipynb](https://github.com/abetlen/llama-cpp-python/blob/main/examples/notebooks/Functions.ipynb) I am getting the following error on the server side.

```
Exception: 'dict object' has no attribute 'name'
  File "/opt/app-root/lib64/python3.11/site-packages/llama_cpp/server/errors.py", line 171, in custom_route_handler
    response = await original_route_handler(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/fastapi/routing.py", line 278, in app
    raw_response = await run_endpoint_function(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/fastapi/routing.py", line 191, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/llama_cpp/server/app.py", line 483, in create_chat_completion
    ] = await run_in_threadpool(llama.create_chat_completion, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/starlette/concurrency.py", line 42, in run_in_threadpool
    return await anyio.to_thread.run_sync(func, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/anyio/to_thread.py", line 56, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/anyio/_backends/_asyncio.py", line 2177, in run_sync_in_worker_thread
    return await future
           ^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/anyio/_backends/_asyncio.py", line 859, in run
    result = context.run(func, *args)
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/llama_cpp/llama.py", line 1748, in create_chat_completion
    return handler(
           ^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/llama_cpp/llama_chat_format.py", line 2610, in functionary_new_chat_handler
    prompt = prepare_messages_for_inference(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/llama_cpp/llama_chat_format.py", line 2598, in prepare_messages_for_inference
    tokenizer.hf_tokenizer.apply_chat_template(all_messages, add_generation_prompt=True, tokenize=False)
  File "/opt/app-root/lib64/python3.11/site-packages/transformers/tokenization_utils_base.py", line 1812, in apply_chat_template
    rendered_chat = compiled_template.render(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/jinja2/environment.py", line 1304, in render
    self.environment.handle_exception()
  File "/opt/app-root/lib64/python3.11/site-packages/jinja2/environment.py", line 939, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "<template>", line 9, in top-level template code
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'name'
```

## Debugger

Running in debugger the function calling example, the first call works perfectly, I am getting the expected response with the right formatting.

However, the second call, where we add the tool response to the messages, got the error above.

Here is the `messages` content before sending it for the second call.

```python
[
  {
    'role': 'user', '
    content': "what's the weather like in Hanoi?"
  }, 
  ChatCompletionMessage(
    content=None, 
    role='assistant', 
    function_call=None, 
    tool_calls=[
    ChatCompletionMessageToolCall(
      id='call_L4tj8aG7VYqlL1XCLzyHeH8z', 
      function=Function(
        arguments='{"location": "Hanoi"}', 
        name='get_current_weather'), 
        type='function'
      )
    ]
  ), 
  {
  'tool_call_id': 'call_L4tj8aG7VYqlL1XCLzyHeH8z', 
  'role': 'tool', 
  'name': 'get_current_weather', 
  'content': '{"location": "Hanoi", "temperature": "unknown"}'
  }
]
```

## Reproduce

To reproduce without using the jupyter, here is the python
```python
import json
import openai

BASE_URL = "http://0.0.0.0:8000/v1"
API_KEY = "sk-xxx"
MODEL_ID = "/models/functionary-small-v2.5.Q4_0.gguf"

def get_current_weather(location, unit="fahrenheit"):
    return json.dumps({"location": location, "temperature": "unknown"})

client = openai.OpenAI(base_url=BASE_URL, api_key=API_KEY)

messages = [
    {"role": "user", "content": "what's the weather like in Hanoi?"}
]
tools = [ # For functionary-7b-v2 we use "tools"; for functionary-7b-v1.4 we use "functions" = [{"name": "get_current_weather", "description":..., "parameters": ....}]
    {
        "type": "function",
        "function": {
            "name": "get_current_weather",
            "description": "Get the current weather",
            "parameters": {
                "type": "object",
                "properties": {
                    "location": {
                        "type": "string",
                        "description": "The city and state, e.g., San Francisco, CA",
                    }
                },
                "required": ["location"]
            }
        }
    }
]

response = client.chat.completions.create(
    model=MODEL_ID,
    messages=messages,
    tools=tools,
)

response_message = response.choices[0].message
tool_calls = response_message.tool_calls

if not tool_calls:
    exit(1)

available_functions = {
            "get_current_weather": get_current_weather,
        }

messages.append(response_message)

for tool_call in tool_calls:
    function_name = tool_call.function.name
    function_to_call = available_functions[function_name]
    function_args = json.loads(tool_call.function.arguments)
    function_response = function_to_call(
        location=function_args.get("location"),
        unit=function_args.get("unit"),
    )
    messages.append(
        {
            "tool_call_id": tool_call.id,
            "role": "tool",
            "name": function_name,
            "content": function_response,
        }
    )  # extend conversation with function response
second_response = client.chat.completions.create(
    model=MODEL_ID,
    messages=messages,
)
```

llama_cpp/llama_chat_format.py
```diff
@@ -1826,7 +2544,6 @@ def generate_schema_from_functions(functions, namespace="functions") -> str:
     def prepare_messages_for_inference(
         messages: List[llama_types.ChatCompletionRequestMessage],
```
axel7083, line 2545: For very weird reason, the messages do not contain the `name` property when the role is `tool`.

Here, in the example I provided https://github.com/abetlen/llama-cpp-python/pull/1509#pullrequestreview-2215230004 the `messages` is equal to 

```json
[
....
   {
      "role":"tool",
      "content":"{\"location\": \"Hanoi, Vietnam\", \"temperature\": \"unknown\"}",
      "tool_call_id":"call_yQZk3sIQsIwGuco1lV4QVbTN"
   }
]
```

In the tool message we do not have the `name` which is required by the chat_format. leading to the error `'dict·object'·has·no·attribute·'name'` raised by jinja.

llama_cpp/llama_chat_format.py
```diff
@@ -1328,7 +1329,7 @@ def format_gemma(
 # Tricky chat formats that require custom chat handlers
 
 
-@register_chat_completion_handler("functionary")
+@register_chat_completion_handler("functionary-old")
```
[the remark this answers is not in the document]
axel7083: maybe `functionary-legacy` is a better name ?

PR #1513, opened 2024-06-06
Add methods to explicitly free model from memory

jkawamoto:
This PR introduces a `close` method to `Llama` and `_LlamaModel`, allowing users to explicitly free the model from RAM/VRAM.

The previous implementation relied on the destructor of `_LlamaModel` to free the model. However, in Python, the timing of destructor calls is unclear: for instance, the `del` statement does not guarantee immediate invocation of the destructor.

This PR offers an explicit method to release the model, which works immediately and enables the user to load another model without encountering memory issues. Here is an example:

```python
from llama_cpp import Llama

llama = Llama(...)
...
llama.close()
```

Additionally, this PR allows the the `Llama` class class to be used as a context manager via the `contextlib.closing` interface. Here is an example:

```python
from contextlib import closing
from llama_cpp import Llama

with closing(Llama(...)) as llama:
    ...
    # the model will be freed here

...
```

### DIFF
--- a/llama_cpp/_internals.py
+++ b/llama_cpp/_internals.py
@@ -1,12 +1,15 @@
 from __future__ import annotations
 
+import contextlib
 import os
 import ctypes
+from types import TracebackType
 
 from typing import (
     List,
     Optional,
     Sequence,
+    Type,
 )
 from dataclasses import dataclass, field
 
@@ -23,7 +26,7 @@ import llama_cpp.llama_cpp as llama_cpp
 # Python wrappers over llama.h structs
 
 
-class _LlamaModel:
+class _LlamaModel(contextlib.AbstractContextManager):
     """Intermediate Python wrapper for a llama.cpp llama_model.
     NOTE: For stability it's recommended you use the Llama class instead."""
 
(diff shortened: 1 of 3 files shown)


abetlen:
@jkawamoto definitely agree here, thank you for the contribution.

A few things I noticed:
- We may need to perform some runtime checks / asserts to make sure the model hasn't been closed
- Right now you're only closing the model explicitly but the context and batch (which both also hold memory) still use the`__del__`

I also think we should have all the internal classess support the `close` / context manager approach, then we can use `contextlib.ExitStack` in the main `Llama` class to ensure they're cleaned up properly.

jkawamoto:
@abetlen Thanks for your comment.

I implemented a context manager in three internal classes and added an `ExitStack` to the `Llama` class.

Regarding the runtime check of whether the model, context, and batch are still loaded, it appears that the three inner classes verify the low-level object isn't `None` before freeing the object. Thus, we can safely close these objects more than once. Would it be necessary to add any additional checks?

abetlen:
@jkawamoto I've update the use of ExitStacks slightly, additionally I removed the explicit `__enter__` / `__exit__` methods. Since we now have a `.close()` method we can use `contextlib.closing` from the standard library to turn the Llama class into a context manager automatically.

```python
from contextlib import closing
from llama_cpp import Llama

with closing(Llama(...)) as llama:
    ...
    # the model will be freed here
```

jkawamoto:
@abetlen I think it would be better if the `Llama` class implements a context manager, as this would simplify the initialization of an instance by requiring fewer input characters.

Additionally, it is still worthwhile to retain the `__del__` method in `Llama` to prevent memory leaks in cases where the user forgets to close the object.

PR #1515, opened 2024-06-06
Workflow update - PART 2

Smartappli:
Add CPU with AVX, AVX2, AVX512 with OpenBlas & Remove unnecesary 32 bits wheels

    Without AVX Utuntu, Windows => 32 bits, mac => 64 bits
    AVX : Ubuntu, Windows, Mac => 32/64 bits
    AVX2: Ubuntu, Windows, Mac => 64 bits
    AVX512 Ubuntu, Windows, Mac => 64 bits


Smartappli:
https://github.com/abetlen/llama-cpp-python/pull/1526

PR #1526, opened 2024-06-13
Workflow speed up - POST PART 1

Smartappli:


abetlen:
@Smartappli I was looking at this a few months ago as well because uv is a pretty amazing tool. The issue I ran into however is that it doesn't log the output of subcommands (ie cmake), so if the build fails it's kinda tough to debug.

[Related Issue](https://github.com/astral-sh/uv/issues/1567) and [PR](https://github.com/astral-sh/uv/pull/2156)

Smartappli:
@abetlen Thank you for your advice. I think I have solved this problem.

abetlen:
I tried setting `RUST_LOG=trace` locally but I still don't see the actual logs from the cmake build system.

Smartappli:
Ping @abetlen

abetlen (approved):

abetlen:
@Smartappli thank you for your contribution and patience on this one, merging now that uv has support for the build logs with `--verbose`

PR #1530, opened 2024-06-14
Dockerfile update

Smartappli:


Smartappli:
@abetlen i think it would be nice to publish docker images on docker hub too

abetlen:
@Smartappli thank you for the contribution, I've removed the docker login for now because I feel that ghcr is sufficient but may reconsider in the future.

PR #1550, opened 2024-06-24
Change server approach to handle parallel requests

sergey-zinchenko:
I have made a change to the way the server handles concurrent requests. In this PR, arriving requests will wait for the model's global async lock. I.e., requests will be organized in something like a queue. On top of that, I added a configuration for the unicorn to have only ten concurrent requests. So finally, up to ten parallel requests will await like "in a queue" for the model lock, and the current request will not be interrupted. If 11's request arrives,  the server will send out 503 response immediately. This approach suits the common scenarios with multiuser chatbot UI and API access.  
I also changed some other stuff to fix PEP warnings by linter in IDE.

### DIFF
--- a/llama_cpp/server/__main__.py
+++ b/llama_cpp/server/__main__.py
@@ -74,6 +74,7 @@ def main():
         port=int(os.getenv("PORT", server_settings.port)),
         ssl_keyfile=server_settings.ssl_keyfile,
         ssl_certfile=server_settings.ssl_certfile,
+        limit_concurrency=10
     )
 
 
--- a/llama_cpp/server/__main__.py
+++ b/llama_cpp/server/__main__.py
@@ -34,7 +34,7 @@ from llama_cpp.server.settings import (
     Settings,
     ServerSettings,
     ModelSettings,
-    ConfigFileSettings,
+    read_config,
 )
 from llama_cpp.server.cli import add_args_from_model, parse_model_from_args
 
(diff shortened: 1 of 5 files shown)


sergey-zinchenko:
@abetlen What do you think about that changes?

Smartappli:
LGTM

Smartappli (approved):

gerdemann:
Hey, thanks for this pr. Is it possible that we can get the pr merged? 😄

sergey-zinchenko:
@gerdemann @Smartappli and I se some activity during that two month related to the way how server handles parallel requests in main branch. Is that still an issue?

gerdemann:
I still get this error when two requests are made at the same time:

```
disconnected
Disconnected from client (via refresh/close) Address(host='10.32.20.82', port=58506)
ERROR:    ASGI callable returned without completing response.
Llama.generate: 64 prefix-match hit, remaining 45 prompt tokens to eval

Fu-Cheng:
> RuntimeError: Task <Task pending name='Task-7' coro=<RequestResponseCycle.run_asgi() running at /llama/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py:406> cb=[set.discard()]> got Future <Future pending> attached to a different loop
> INFO:     10.32.20.82:37322 - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error

@sergey-zinchenko, 
Hi, I encountered the same issue. I found that setting --embedding False helps to bypass the error. However, the service is still not handling concurrent requests properly. When I send a second request while the LLM is still generating a response for the first request, I receive the following error in my frontend:

HTTP 400 Bad Request from http://localhost:8080/v1/chat/completions

Fu-Cheng:
> RuntimeError: Task <Task pending name='Task-7' coro=<RequestResponseCycle.run_asgi() running at /llama/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py:406> cb=[set.discard()]> got Future <Future pending> attached to a different loop
> INFO:     10.32.20.82:37322 - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error

@sergey-zinchenko,

Hi, I encountered the same issue. The service is still not handling concurrent requests properly. When I send a second request while the LLM is still generating a response for the first request, I receive this error.

gjpower:
I have implemented a smaller alternative change that solves the same problem in #1798

PR #1551, opened 2024-06-25
Fix the CUDA workflow

### LINKED ISSUE #1543: CUDA wheels are not being built by GitHub actions
Currently, the workflow that builds CUDA wheels is failing with the following error:

> C:\Miniconda3\envs\build\include\crt/host_config.h(153): fatal error C1189: #error:  -- unsupported Microsoft Visual Studio version! Only the versions between 2017 and 2022 (inclusive) are supported! The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk. [C:\Users\runneradmin\AppData\Local\Temp\tmpwbsbwtdg\build\CMakeFiles\CMakeScratch\TryCompile-uh6ciq\cmTC_cbbed.vcxproj]

See:

* https://github.com/abetlen/llama-cpp-python/actions/runs/9457182450/job/26051175939
* https://github.com/oobabooga/llama-cpp-python-cuBLAS-wheels/actions/runs/9457447475/job/26051277254.

The relevant workflow is [build-wheels-cuda.yaml](https://github.com/abetlen/llama-cpp-python/blob/main/.github/workflows/build-wheels-cuda.yaml), which is based on the original workflows by @jllllll [in this repository](https://github.com/jllllll/llama-cpp-python-cuBLAS-wheels) (no longer maintained).

-----

I have tried adding the `-allow-unsupported-compiler` flag to my own workflows, as suggested in the error message, and while that did allow the wheel to be successfully compiled, it crashed on my Windows tests. A better solution would be to set a speficic past Visual Studio version in the workflow to avoid similar errors in the future.

This ExLlamaV2 workflow does that, but I have been unable to port it to the llama-cpp-python workflow: [build-wheels-release.yml](https://github.com/turboderp/exllamav2/blob/5996922a0f0937aa503efa773780f1648915d73e/.github/workflows/build-wheels-release.yml#L120).

oobabooga:
This pins Visual Studio and Windows Server to specific past versions, avoiding a recent update that broke compatibility with CUDA 12.1.

Confirmed to be working on my llama-cpp-python-cuBLAS-wheels fork: https://github.com/oobabooga/llama-cpp-python-cuBLAS-wheels/actions/runs/9651433295/job/26619287975

Closes #1543

### DIFF
--- a/.github/workflows/build-wheels-cuda.yaml
+++ b/.github/workflows/build-wheels-cuda.yaml
@@ -20,7 +20,7 @@ jobs:
         id: set-matrix
         run: |
           $matrix = @{
-              'os' = @('ubuntu-latest', 'windows-latest')
+              'os' = @('ubuntu-latest', 'windows-2019')
               'pyver' = @("3.9", "3.10", "3.11", "3.12")
               'cuda' = @("12.1.1", "12.2.2", "12.3.2", "12.4.1")
               'releasetag' = @("basic")
@@ -43,6 +43,12 @@ jobs:
       AVXVER: ${{ matrix.releasetag }}
 
     steps:
+      - name: Add MSBuild to PATH
+        if: runner.os == 'Windows'
+        uses: microsoft/setup-msbuild@v1.1
+        with:
+          vs-version: '[16.11,16.12)'
+
       - uses: actions/checkout@v4
         with:
           submodules: "recursive"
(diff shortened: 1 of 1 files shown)


Smartappli (approved):

gaby (approved):

gaby:
Ping @abetlen

abetlen:
@oobabooga thank you so much for this! Been travelling for the last week, should have a new release tonight!

Issue #1498, opened 2024-05-30
Include usage key in create_completion when streaming


zhudotexe:
**Is your feature request related to a problem? Please describe.**
Since `create_completion` may yield text chunks comprised of multiple tokens per yield (e.g. in the case of multi-byte Unicode characters), counting the number of yields may not equal the number of tokens actually generated by a model. To accurately get the usage statistics of a streamed completion, one has to run the final text through the tokenizer again, despite `create_completion` already tracking the number of tokens generated by the model.

**Describe the solution you'd like**
When `stream=True` in `create_completion`, the final chunk yielded should include the usage statistics in the `'usage'` key.

**Describe alternatives you've considered**
- Saving full generated text and running it through the tokenizer again (seems wasteful)
- Counting the number of yields and hoping we don't have any multi-byte characters (hacky and fragile)

**Additional context**
The OpenAI API has recently added similar support in their streaming API with the `stream_options` key: https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options

hh23485:
any updates on this? Seems that very helpful for our situation as well as the committer' desc

hh23485:
Any updates?

PR #1552, opened 2024-06-25
This commit adds streaming options

tpfau:
This PR adds the `stream_options` and `include_usage` fields as defined in the OpenAI API reference to the server.
Set to false by default they can be activated which leads to the last chunk being returned during streaming to supply the usage information. 
This solves the Question raised in #1461 and the request indicated in #1498.
I decided to put the chunk generation code into its own function since not doing so would have lead to blown up code. 
I did not update the places where `stream` is clearly set to `false`. 

I'm happy to update the PR with any necessary/requested modifications wrt code style or if you think all chunk generation places should be changed to use functions (or add more options/update the chunk generation function). Just let me know.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -946,6 +946,56 @@ class Llama:
         else:
             return output
 
+    def _create_chunk(
+        self,
+        completion_id: str,
+        created: int,
+        model_name: str,
+        text: str,
+        logprobs_or_none: Union[Optional[CompletionLogprobs], None],
+        include_usage: bool,
+        index: int,
+        finish_reason: Union[str, None],
+        usage: Union[Dict[str, Any], None] = None,
+    ) -> CreateChatCompletionStreamResponse:
+        """
+        Create chunks for streaming API, depending on whether usage is requested or 
+        not they need (or don't need) an additional field
+        """
+
+        if include_usage:
+            token = {
+                "id": completion_id,
+                "object": "text_completion",
+                "created": created,
+                "model": model_name,
+                "choices": [
+                    {
+                        "text": text,
+                        "index": index,
+                        "logprobs": logprobs_or_none,
+                        "finish_reason": finish_reason,
+                    },
+                ],
+                "usage": usage,
+            }
+        else:
+            token = {
+                "id": completion_id,
+                "object": "text_completion",
+                "created": created,
+                "model": model_name,
+                "choices": [
+                    {
+                        "text": text,
+                        "index": index,
+                        "logprobs": logprobs_or_none,
+                        "finish_reason": finish_reason,
+                    }
+                ],
+            }
+        return token
+
     def _create_completion(
         self,
         prompt: Union[str, List[int]],
@@ -963,6 +1013,7 @@ class Llama:
         repeat_penalty: float = 1.1,
         top_k: int = 40,
         stream: bool = False,
+        stream_include_usage: Optional[bool] = False,
         seed: Optional[int] = None,
         tfs_z: float = 1.0,
         mirostat_mode: int = 0,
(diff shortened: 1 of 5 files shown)


abetlen:
@tpfau thank you for starting on this, I'll review in more depth but my initial request would be that instead of `stream_include_usage` we just add `stream_options` directly to the methods in the `Llama` class as they should also mirror the OpenAI api.

tpfau:
@abetlen I hope this change is what you had in mind. 
I'm admittedly unsure how I can properly add the description of the `include_usage` field description to the API (I tried to stick to how it was done with the completion requests specs for now). 
I also wanted to avoid having to add the stream option check in every single function so I kept it in the "main" functionalities and left the addition of the `stream_include_usage` parameter in the conversion functions.

tpfau:
I would really like to see this in production, as I think this feature can be quite important for a lot of applications that need some kind of usage control or just let the users know about their usage. Of course people can work around this kind of approach but it requires additional code in the code base that does tokenization and thus computation which is completely unnecessary as it is already calculated by the model.

tpfau:
@abetlen Any news on this?

tpfau:
Updated again to fix conflicts. 

Would be great to see this feature in the codebase. If there is anything missing, please let me know

tpfau:
Did some tests again, seems to be working. Please let me know if you have any updates which you want to have in.

llama_cpp/llama.py
```diff
@@ -1363,6 +1414,11 @@ def logit_bias_processor(
                 break
 
             if stream:
+                if stream_options is not None and "include_usage" in stream_options:
+                    include_usage = True if stream_options["include_usage"] else False
+                else:
+                    include_usage = False
```
lukestanley, line 1420: Instead of the double nested if block, presumably you could do: `if stream and stream_options and "include_usage" in stream_options:`? @tpfau

llama_cpp/server/app.py
```diff
@@ -495,7 +496,7 @@ async def create_chat_completion(
             if body.logit_bias_type == "tokens"
             else body.logit_bias
         )
-
+
```
lukestanley, line 499: Does Black or Ruff need to be ran? @tpfau

llama_cpp/server/types.py
```diff
@@ -228,6 +237,10 @@ class CreateChatCompletionRequest(BaseModel):
     min_p: float = min_p_field
     stop: Optional[Union[str, List[str]]] = stop_field
     stream: bool = stream_field
+    stream_options: Optional[llama_cpp.StreamOptions] = Field(
+        default=None,
+        description="Options for streaming response. Only set this when you set stream: true.",
+    )
```
lukestanley, line 243: Presumably this is where you would make use of the description you have in `include_usage_field` somehow?
I wonder if [FastAPI dependency injection](https://fastapi.tiangolo.com/tutorial/dependencies/) could help with this, or if there is an example pattern in the server code for providing property help descriptions that is better followed instead. I've not used FastAPI zealously enough to know yet.
@tpfau

llama_cpp/llama.py
tpfau: No you can't since all the following code has to be run in both instances, and only the include stream usage needs to be adapted based on the props.

llama_cpp/server/app.py
tpfau: true

llama_cpp/server/types.py
tpfau: I'm not sure how to do this neither. The problem I have is that llama_types defines `TypedDicts` which don't have descriptions like `BaseModels`. So if anyone got a good suggestion I'm happy to update this.

llama_cpp/llama.py
lukestanley: Ahh, of course. I guess I was meant something like this:
```python
if stream_options and "include_usage" in stream_options:
    include_usage = stream_options["include_usage"]
else:
    include_usage = False
```
It relies on a None `stream_options` being _falsey_, and `stream_options.include_usage` being a boolean.
Just a little more readable.

PR #1594, opened 2024-07-12
Actually use split_mode from model settings

grider-withourai:
One of the parameters, `split_mode`, was not used when loading. Fix it by passing it to the Llama server.


grider-withourai:
Tested working with some models.

abetlen:
@grider-withourai thank you for catching this!

PR #1605, opened 2024-07-17
Updated README.md, llama_cpp/llama.py and pyproject.toml to add support for cross-encoders

perpendicularai:


perpendicularai:
See video :
https://github.com/user-attachments/assets/de8a0d3b-9c75-4a60-928d-389bb8bdea2d

PR #1635, opened 2024-07-29
FreeBSD compatibility

yurivict:


abetlen:
@yurivict fantastic, thank you for the contribution!

PR #1637, opened 2024-07-29
Ported back new grammar changes from C++ to Python implementation

### LINKED ISSUE #1547: Grammars bracket repetition symbol not working
Hello, I tried checking for similar issues about this problem but couldn't find one.
I've had an issue with not being able to use the repetition brackets symbol when working with grammars.

I'm using Ubuntu 20.04, Python 3.12 and `llama_cpp_python==0.2.79`.

The following works fine:

```python
from llama_cpp import LlamaGrammar

grammar_string = r"""root ::= "repeating" [a-z]+"""
my_grammar = LlamaGrammar.from_string(grammar_string, verbose=True)
```

But this doesn't:
```python
from llama_cpp import LlamaGrammar

grammar_string = r"""root ::= "repeating" [a-z]{1,}"""
my_grammar = LlamaGrammar.from_string(grammar_string, verbose=True)
```

It returns this error:
```
parse: error parsing grammar: expecting newline or end at {1,}
    my_grammar = LlamaGrammar.from_string(grammar_string, verbose=True)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/anaconda3/envs/DocLLM/lib/python3.12/site-packages/llama_cpp/llama_grammar.py", line 71, in from_string
    raise ValueError(
ValueError: from_string: error parsing grammar file: parsed_grammar.rules is empty
```
The llama.cpp [GBNF Guide](https://github.com/ggerganov/llama.cpp/blob/master/grammars/README.md#repetition-and-optional-symbols) seems to say it should be possible to use this pattern?

> # Repetition and Optional Symbols
> 
>     * after a symbol or sequence means that it can be repeated zero or more times (equivalent to {0,}).
>     + denotes that the symbol or sequence should appear one or more times (equivalent to {1,}).
>     ? makes the preceding symbol or sequence optional (equivalent to {0,1}).
>     {m} repeats the precedent symbol or sequence exactly m times
>     {m,} repeats the precedent symbol or sequence at least m times
>     {m,n} repeats the precedent symbol or sequence at between m and n times (included)
>     {0,n} repeats the precedent symbol or sequence at most n times (included)

Not sure if my understanding of GBNF is lacking or if it's a real bug.
Thank you!

ExtReMLapin:
https://github.com/ggerganov/llama.cpp/pull/6640
https://github.com/ggerganov/llama.cpp/pull/6467
https://github.com/ggerganov/llama.cpp/pull/7194

Fixes #1547

### DIFF
--- a/llama_cpp/llama_cpp.py
+++ b/llama_cpp/llama_cpp.py
@@ -3002,7 +3002,7 @@ def llama_grammar_init(
     n_rules: Union[ctypes.c_size_t, int],
     start_rule_index: Union[ctypes.c_size_t, int],
     /,
-) -> llama_grammar_p:
+) -> Optional[llama_grammar_p]:
     """Initialize a grammar from a set of rules."""
     ...
 
--- a/llama_cpp/llama_grammar.py
+++ b/llama_cpp/llama_grammar.py
@@ -432,10 +432,12 @@ class std:
 #     // be an inclusive range ([a-z])
 #     LLAMA_GRETYPE_CHAR_RNG_UPPER = 5,
 
-
 #     // modifies a preceding LLAMA_GRETYPE_CHAR or
 #     // LLAMA_GRETYPE_CHAR_RNG_UPPER to add an alternate char to match ([ab], [a-zA])
 #     LLAMA_GRETYPE_CHAR_ALT       = 6,
+
+#     // any character (.)
+#     LLAMA_GRETYPE_CHAR_ANY       = 7,
 # };
 class llama_gretype(Enum):
     """grammar element type"""
@@ -447,6 +449,7 @@ class llama_gretype(Enum):
     LLAMA_GRETYPE_CHAR_NOT = 4  # inverse char(s) ([^a], [^a-b] [^abc])
     LLAMA_GRETYPE_CHAR_RNG_UPPER = 5  # modifies a preceding LLAMA_GRETYPE_CHAR or LLAMA_GRETYPE_CHAR_ALT to be an inclusive range ([a-z])
     LLAMA_GRETYPE_CHAR_ALT = 6  # modifies a preceding LLAMA_GRETYPE_CHAR or LLAMA_GRETYPE_CHAR_RNG_UPPER to add an alternate char to match ([ab], [a-zA])
+    LLAMA_GRETYPE_CHAR_ANY = 7  # any character (.)
 
 
 # struct parse_state {
@@ -830,6 +833,10 @@ def parse_sequence(
         #     if (last_sym_start == out_elements.size()) {
         #         throw std::runtime_error(std::string("expecting preceeding item to */+/? at ") + pos);
         #     }
+        elif pos[0] == '.':
+            last_sym_start = out_elements.size()
+            out_elements.push_back(LlamaGrammarElement(llama_gretype.LLAMA_GRETYPE_CHAR_ANY, 0))
+            pos = parse_space(pos + 1, is_nested)
         elif pos[0] in ("*", "+", "?"):  # repetition operator
             if last_sym_start == out_elements.size():
                 raise RuntimeError("expecting preceding item to */+/? at " + str(pos))
@@ -1039,6 +1046,7 @@ def is_char_element(elem: LlamaGrammarElement) -> bool:
         llama_gretype.LLAMA_GRETYPE_CHAR_NOT,
         llama_gretype.LLAMA_GRETYPE_CHAR_ALT,
         llama_gretype.LLAMA_GRETYPE_CHAR_RNG_UPPER,
+        llama_gretype.LLAMA_GRETYPE_CHAR_ANY,
     )
 
 
@@ -1135,6 +1143,8 @@ def print_rule(
                     + str(i)
                 )
             print_grammar_char(file, elem.value)
+        elif case is llama_gretype.LLAMA_GRETYPE_CHAR_ANY:
+            print(".", file=file, end="")
         # if (is_char_element(elem)) {
         #     switch (rule[i + 1].type) {
         #         case LLAMA_GRETYPE_CHAR_ALT:
--- a/llama_cpp/llama_grammar.py
+++ b/llama_cpp/llama_grammar.py
@@ -1,6 +1,12 @@
 """Python implementation of llama grammar parser directly translated from C++ source file in vendor/llama.cpp/common/grammar-parser.cpp."""
 
 # flake8: noqa
+import sys
+import ctypes
+import enum
+import typing
+import dataclasses
+
 from itertools import groupby
 from typing import (
     Any,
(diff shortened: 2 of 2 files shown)


ExtReMLapin:
Not working yet, for example : 


Ok  :  `root  ::=  ("EYEYAHA"){5}`


Not working : `root  ::=  ("EYEYAHA"){1,5}`

```txt

from_string grammar:
root ::= root_1 root_5
root_1 ::= [E] [Y] [E] [Y] [A] [H] [A]
root_2 ::= root_1 | print_grammar: error printing grammar: unexpected end of rule: 2,2

```

ExtReMLapin:
I've been looking again and again and I don't see what I missed from the pr diff 😕

ExtReMLapin:
Help is welcome if you can help @abetlen

abetlen:
Hey @ExtReMLapin thanks for starting on this fix, just getting back to everything now after vacation. I'll take a stab at this over the next couple days as well.

abetlen:
@ExtReMLapin got the new grammar features back-ported and ended up rewriting most of `llama_grammar.py`.  With #1649 this should bring the grammar implementation in-line with llama.cpp.

ExtReMLapin:
Thank you abetlen.
While checking the code I was a little surprised by the multiple else-if and the list hardcoded inside the function instead of being outside (so regenerated on each function call) (ex : decode_utf8)


As a proud lazy man, I asked GPT4 to 

1. Try to write an optimized version of `parse_hex` `decode_utf8` `parse_char` (I actually expected it to make a jmp table for it)
2. Write a benchmark and tests for it

Issue is that is seems that not all UTF8 characters are supported
(see test code bellow), Is it really an issue ?



```python

import timeit
import typing

# Original Functions
def original_parse_hex(src: str, size: int) -> typing.Tuple[int, str]:
    pos = 0
    value = 0
    for _ in range(size):
        value <<= 4
        c = src[pos]
        if "a" <= c <= "f":
            value += ord(c) - ord("a") + 10
        elif "A" <= c <= "F":
            value += ord(c) - ord("A") + 10
        elif "0" <= c <= "9":
            value += ord(c) - ord("0")
        else:
            break
        pos += 1
    if pos != size:
        raise ValueError(f"expecting {size} hex chars at {src}")
    return value, src[pos:]


def original_decode_utf8(src: str) -> typing.Tuple[int, str]:
    lookup: list[int] = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 3, 4]
    first_byte: int = ord(src[0])
    highbits: int = first_byte >> 4
    #first_byte to hex 

    print(highbits)
    length: int = lookup[highbits]
    mask: int = (1 << (8 - length)) - 1
    value: int = first_byte & mask
    end: int = min(len(src), length)

    pos: int = 1
    for pos in range(1, end):
        if not src[pos]:
            break
        value = (value << 6) + (ord(src[pos]) & 0x3F)

    return value, src[pos:] if pos < len(src) else ""


def original_parse_char(src: str) -> typing.Tuple[int, str]:
    if src[0] == "\\":
        if src[1] == "x":
            return original_parse_hex(src[2:], 2)
        elif src[1] == "u":
            return original_parse_hex(src[2:], 4)
        elif src[1] == "U":
            return original_parse_hex(src[2:], 8)
        elif src[1] == "t":
            return ord("\t"), src[2:]
        elif src[1] == "r":
            return ord("\r"), src[2:]
        elif src[1] == "n":
            return ord("\n"), src[2:]
        elif src[1] in ('\\', '"', '[', ']'):
            return ord(src[1]), src[2:]
        else:
            raise ValueError(f"unknown escape at {src}")
    elif src:
        return original_decode_utf8(src)
    raise ValueError("unexpected end of input")


hex_map = {**{f"{x}": x for x in range(10)}, **{chr(x): x - ord('a') + 10 for x in range(ord('a'), ord('f') + 1)}, **{chr(x): x - ord('A') + 10 for x in range(ord('A'), ord('F') + 1)}}
# Optimized Functions
def optimized_parse_hex(src: str, size: int) -> typing.Tuple[int, str]:
    
    value = 0
    for i in range(size):
        c = src[i]
        if c in hex_map:
            value = (value << 4) + hex_map[c]
        else:
            raise ValueError(f"expecting {size} hex chars at {src}")
    return value, src[size:]


prealloc = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 3, 4]

def optimized_decode_utf8(src: str) -> typing.Tuple[int, str]:
    first_byte = ord(src[0])
    highbits = first_byte >> 4
    length = prealloc[highbits]
    value = first_byte & ((1 << (8 - length)) - 1)

    for i in range(1, length):
        value = (value << 6) + (ord(src[i]) & 0x3F)
    
    return value, src[length:]


escape_sequences = {
    "x": 2, "u": 4, "U": 8,
    "t": ord("\t"), "r": ord("\r"), "n": ord("\n"),
    "\\": ord("\\"), '"': ord('"'), '[': ord('['), ']': ord(']')
}

def optimized_parse_char(src: str) -> typing.Tuple[int, str]:
    if src[0] == "\\":

        esc = src[1]
        if esc in escape_sequences:
            if esc in 'xuU':
                return optimized_parse_hex(src[2:], escape_sequences[esc])
            return escape_sequences[esc], src[2:]
        raise ValueError(f"unknown escape at {src}")
    elif src:
        return optimized_decode_utf8(src)
    raise ValueError("unexpected end of input")

import random 
def generate_utf8_string(length: int) -> str:
    utf8_chars = [
        chr(random.randint(0x20, 0x7E)),    # ASCII characters
        chr(random.randint(0x80, 0x07FF)),  # Extended Latin and similar
        chr(random.randint(0x0800, 0xFFFF)),  # Multilingual Plane
        chr(random.randint(0x10000, 0x10FFFF)) # Supplementary Planes (Emoji, etc.)
    ]
    return ''.join(random.choice(utf8_chars) for _ in range(length))

def benchmark():
    # Generate a random UTF-8 string of 500 characters
    test_string = generate_utf8_string(500)
    print('Random string : ', test_string)
    
    # Ensure both functions return the same result
    original_result = original_parse_char(test_string)
    optimized_result = optimized_parse_char(test_string)
    
    assert original_result == optimized_result, "The results of original and optimized functions do not match!"
    
    original_time = timeit.timeit(lambda: original_parse_char(test_string), number=100000)
    optimized_time = timeit.timeit(lambda: optimized_parse_char(test_string), number=100000)
    
    print(f"Original parse_char time: {original_time:.6f} seconds")
    print(f"Optimized parse_char time: {optimized_time:.6f} seconds")

if __name__ == "__main__":
    benchmark()
```

ExtReMLapin:
Alright, I gave a try at the office, rule parsing is broken, and few functions are missing (ex `from_file`), i'll take a look

abetlen:
@ExtReMLapin just fixed the last bug, was re-assigning `out_elements` by mistake inside of `parse_sequence`.

ExtReMLapin:
Thanks for the fix, will do !

PR #1667, opened 2024-08-07
Remove spurious bracket added in PR 1656

CISC:
Remove an unintentionally added parenthesis in #1656 that broke the build.


abetlen:
@CISC thank you, just saw this after fixing the same thing.

CISC:
@abetlen No worries. :)

PR #1677, opened 2024-08-13
Allow server to accept openai's new structured output "json_schema" format.

cerealbox:
convert openai response_format: 
`{ type: "json_schema" }`
to the response_format llama-cpp-python expects:
`{ type: "json_object", "schema": ... }`

is this how this should be done?


aqni:
This conversion would be really useful for OpenAI API compatibility — hoping the maintainer can review and merge it!

PR #1704, opened 2024-08-25
fix: Fix memory allocation of ndarray

xu-song:
## Motivation

https://github.com/abetlen/llama-cpp-python/blob/259ee151da9a569f58f6d4979e97cfd5d5bc3ecd/llama_cpp/llama.py#L2052

The above part will allocate a large amount of memory.
This commit reduces the memory consumption from more than 10GB to 0GB.

## Simle Reproduce

```py
import os
import numpy as np
import psutil

def print_mem():
    mem = psutil.Process(os.getpid()).memory_info().rss / 1024 / 1024 / 1024
    print(f"process_mem: {mem:.2f} GB")

n_ctx = 32768
n_vocab = 150000
scores = np.ndarray((n_ctx, n_vocab), dtype=np.single)
print_mem()   # 0.03 GB;

scores[:1000, :] = 0.5
print_mem()  # 0.59 GB

# zero-ing out the rest
scores[1000:, :] = 0.0     #  ----- large memory allocation
print_mem()  # 18.34 GB
```

## After this pr

zero-ing out the rest will not allocate extra memory if not necessary.
```py
# zero-ing out the rest
rest = scores[1000:, :]
rest[rest > 0] = 0.0
print_mem()  # 0.59 GB
```


xu-song:
This bug has not been fixed yet

abetlen:
@xu-song thanks for the contribution. Just added some tests and bug fixes since the recent sampler api changes.

PR #1776, opened 2024-10-01
Update Dockerfile

Smartappli:


lukestanley (commented):
Switching to the new, non-depreciated Ubuntu LTS base for seems like a good idea.
Presumably to test, someone would clone the repo and follow:
https://github.com/abetlen/llama-cpp-python/blob/main/docker/README.md#cuda_simple
Copy / download a GGUF model, then see if they can get LLM generation responses from the server?

PR #1791, opened 2024-10-09
docs: remove double word typo in README

Victoran0:
Under the Chat Completion section, about the fourth line under it, we have: 'The model will will format the messages into a single prompt using the following order of precedence:'... 
'will' appeared twice, this update fix that.


lukestanley (changes requested):
The commit message / PR title could better match what it is! Deleting the typo / grammar issue has nothing to do with cuda.

Victoran0:
> The commit message / PR title could better match what it is! Deleting the typo / grammar issue has nothing to do with cuda.

Oops, sorry about that, I did use the same message as the latest commit the readme.md file has, with the additional (updated) at the end. That is because i thought the new commit is not a new feature, only fixing a typographical error.
However, I have edited the commit message in a new commit.

lukestanley:
But now there are 2 commit messages and none of them quite describe what someone would want to know. For example if someone saw the message: "Typo at line 340 fixed", they would think - what file? Then they might go to check the file and see that there is no change there at all. Empty commits are not so helpful.
Instead, you might want to undo your last empty commit, and adjusting the commit message in your original commit.
You would probably then want to force push it up to overwrite Victoran0:patch-2 with a better commit message.
ChatGPT (among other chatbots) is probably good at helping with this sort of thing.

Victoran0:
@lukestanley how does it look now?

lukestanley (approved):
It's readable, it matches up with what it does. I think it's acceptable. There is no need to say the line number, but it's okay. It's fairly clean and tidy. In my capacity as a random internet person I approve. :+1:  @Victoran0

Victoran0:
Am glad you considered 👏👏👏

PR #1793, opened 2024-10-13
Fix: Refactor Batching notebook to use new sampler chain API

lukestanley:
- Replaced deprecated llama_sample_token with llama_sampler_sample
- Updated llama_token_to_piece signature to include lstrip and special arguments
- Changed llama_load_model_from_file to use positional argument for params (currently does not accept a keyword argument for it)
@abetlen


lukestanley:
I was trying to figure out what regressed with the low level API, and I saw that there was a bit more recent batching notebook. But the Batching notebook itself was broken so I fixed it.

To use GitHub Codespaces after making a Codespace from the branch, I had to:
```
pip install --upgrade pip
pip install -e .
wget "https://huggingface.co/TheBloke/Mistral-7B-v0.1-GGUF/resolve/main/mistral-7b-v0.1.Q2_K.gguf?download=true"
mv "mistral-7b-v0.1.Q2_K.gguf?download=true" "mistral-7b-v0.1.Q2_K.gguf"
```
When Codespace offers to install the Jupyter extension, I accepted it.
I am then able to run the whole notebook.

PR #1795, opened 2024-10-14
Fix: add missing exit_stack.close() to end of /v1/completions endpoint

gjpower:
- /v1/completions endpoint was missing an exit_stack.close() call

This was causing the llama proxy lock to be retained until exit_stack is cleaned up by garbage collector


gjpower:
#1798 supersedes this MR as I found thread starvation issue was not just due to improper closing of exit_stack but due to thread starvation from dependency pull of llama_proxy locking all worker threads

PR #1819, opened 2024-11-02
docs: Remove ref to llama_eval in llama_cpp.py docs

richdougherty:
The `llama_eval` function was removed in 0.2.14 (0d37ce5) but the docs still reference it. Updated some of the docs in `llama_cpp.py` to match the `llama.h` docs from upstream.

_Note: the `examples/low_level_api/*.py` files still reference `llama_eval` and therefore don't work anymore, but I haven't looked into fixing these._


abetlen:
Thank you so much @richdougherty

PR #1838, opened 2024-11-21
Fix: CUDA workflow actions

pabl-o-ce:
Hi guys

I take the time to quick fix the current problem creating the CUDA wheels. remove CUDA version 12.5.0 incompatibility with VS.

There some problems with 12.5.0, 12.5.1, 12.6.3 CUDA versions using the current way to install in windows-2019.

Next step I would keep working to fix on this new versions to make it work properly. Also I migrate to windows-2022 to test the current setup and work but is not included in this PR.


pabl-o-ce:
here is the test that it works: https://github.com/pabl-o-ce/llama-cpp-python/actions/runs/11963297992

PR #1864, opened 2024-12-15
Add: Include option to normalize & truncate embeddings in create_embe…

KanishkNavale:
## What

- Includes option to create **normalized** embeddings with **truncation** while using the "create_embeddings" method.

## Why

Currently, the "create_embeddings" methods does not include this as args. However, embed function includes this. I added this args. in the "create_embeddings" method which parse it to the "embed" method.


SubatomicPlanets:
I'm guessing this isn't merged yet because the pr also includes some formatting changes? In general this would be a nice small quality of life improvement.

KanishkNavale:
@SubatomicPlanets Agreed 💯

PR #1867, opened 2024-12-16
docs: update ROCm install instructions

agronholm:
The updated installation instructions allow the utilization of the GPU, as per the [upstream instructions](https://github.com/ggerganov/llama.cpp/blob/master/docs/build.md#hip).


HardAndHeavy:
Thanks a lot for the correction. It helped me a lot with the installation.

PR #1884, opened 2024-12-25
feat: add streaming tool use

### LINKED ISSUE #1869: chatml-function-callling not adding tool description to the prompt.
Tool description is missing in `chatml-funcion-calling` prompt format.

https://github.com/abetlen/llama-cpp-python/blob/803924b02cc8ca7531b70e6497d1ff7b83f75737/llama_cpp/llama_chat_format.py#L3417

### LINKED ISSUE #1883: Feature request: add support for streaming tool use
The combination `stream=True, tool_choice="auto"` raises an exception right now, which means that developers are stuck with one of two unfortunate choices:
1. Developing an application that streams the response but cannot use tools
2. Developing an LLM application that can use tools but cannot stream the response

Relevant discussion: https://github.com/abetlen/llama-cpp-python/discussions/1615

lsorber:
This PR upgrades the `chatml-function-calling` chat handler with fixes for #1883 and #1869, among other improvements.

Changes:
1. General:
    a. ✨ If no system message is supplied, add an empty system message to hold the tool metadata.
    b. ✨ Add function descriptions to the system message so that tool use is better informed (fixes #1869).
    c. ✨ Replace `print` statements relating to JSON grammars with `RuntimeWarning` warnings.
    d. ✅ Added tests with fairly broad coverage of the different scenarios.
4. Case "Tool choice by user":
    b. ✨ Add support for more than one function call by making this a special case of "Automatic tool choice" with a single tool.
5. Case "Automatic tool choice -> respond with a message":
    a. ✨ Use user-defined `stop` and `max_tokens`.
    b. 🐛 Replace incorrect use of follow-up grammar with user-defined grammar.
6. Case "Automatic tool choice -> one or more function calls":
    a. ✨ Add support for streaming the function calls (fixes #1883).
    b. ✨ Make tool calling more robust by giving the LLM an explicit way to terminate the tool calls by wrapping them in a `<function_calls></function_calls>` block.
    c. 🐛 Add missing ":" stop token to determine whether to continue with another tool call, which prevented parallel function calling.
    d. ✨ Set temperature=0 to determine whether to continue with another tool call, similar to the initial decision on whether to call a tool.

### DIFF
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -7,6 +7,7 @@ import ctypes
 import dataclasses
 import random
 import string
+import warnings
 
 from contextlib import ExitStack
 from typing import (
@@ -28,9 +29,7 @@ from jinja2.sandbox import ImmutableSandboxedEnvironment
 import numpy as np
 import numpy.typing as npt
 
-import llama_cpp.llama as llama
-import llama_cpp.llama_types as llama_types
-import llama_cpp.llama_grammar as llama_grammar
+from llama_cpp import llama, llama_grammar, llama_types
 
 from ._logger import logger
 from ._utils import suppress_stdout_stderr, Singleton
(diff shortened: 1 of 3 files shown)


lsorber:
@abetlen The tests all pass, but the macOS ones were terminated after a timeout. I think this is because of a lack of CPU and or memory resources because the tests run fine on my macOS machine.

SubatomicPlanets:
I would love to see this merged! Actually there are quite a lot of good pull requests here that i would like to see merged... But this one is top priority!

lsorber:
Update: I rebased on the latest `main` and included a few tiny improvements to further improve tool calling robustness.

lsorber:
Update: I rebased on the latest `main` and conditionally skipped the added tests on macOS when not enough resources are available to run them.

conornash:
Would love to see this merged - is there anything holding it up?

lsorber:
@abetlen I rebased the PR on the latest upstream main and added a small commit to fix the returned logprobs format.

tsharp:
I would also like to see this merge. 🧙‍♂️

XyLearningProgramming:
I rebased this PR onto the latest upstream main (v0.3.16) and resolved the merge conflicts. Opened a new PR at #2129 in case it helps get this merged. All credit to @lsorber for the implementation.

PR #1901, opened 2025-01-21
Update llama_cpp: Sync LLAMA_API names with llama.cpp mainline. Needs more testing

JamePeng:
Sync LLAMA_API names with llama.cpp mainline. Needs more testing

PR #1912, opened 2025-01-28
Migrate to the latest version of llama.cpp APIs and support for DeepSeek models

jaepil:
This PR is based on #1901.

By applying this PR, llama-cpp-python can load DeepSeek models without issues.


jaepil:
I didn't know the original author @JamePeng was still working on the PR. It seems like his PR is almost complete, so I'm closing mine.

PR #1966, opened 2025-03-09
Synchronize LLAMA_API with ggml-org/llama.cpp and update cuda workflow for windows

JamePeng:
Update llama.cpp version [llama.cpp](https://github.com/ggml-org/llama.cpp) updated [from 794fe2 to f08f4b3]
Use the llama_sampler_init instead of llama_sampler() for safe usage
Sync  llama : add Phi-4-mini support
Sync  llama : expose llama_model_n_head_kv in the API
Sync  tool-call: fix Qwen 2.5 Coder support, add micro benchmarks, support trigger patterns for lazy grammars
class LlamaSampler: append add_xtc(), add_top_n_sigma() and add_dry()
Remove Tail-Free sampling
Add TopN-Sigma/XTC/DRY samplers code into sampler
Sync llama : Add Gemma 3 support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.3.8]
+
+- feat: Update llama.cpp to ggerganov/llama.cpp@7841fc723e059d1fd9640e5c0ef19050fcc7c698
+
 ## [0.3.7]
 
 - feat: Update llama.cpp to ggerganov/llama.cpp@794fe23f29fb40104975c91fe19f23798f7c726e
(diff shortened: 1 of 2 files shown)


JamePeng:
I tried to adjust the workflow output based on VS2022 to compile pip wheels, and generate two cuda versions 12.4.1 and 12.6.3 and the win version of py310-312 for your convenience.
It should have been compiled now: https://github.com/JamePeng/llama-cpp-python/releases

JamePeng:
llama : refactor llama_context, llama_kv_cache, llm_build_context (https://github.com/ggml-org/llama.cpp/pull/12181)

JamePeng:
The adjusted code is moved to https://github.com/JamePeng/llama-cpp-python/tree/1966-branch

PR #1971, opened 2025-03-15
fix(types): remove redundant type (typo with repeating lines)

CHB-0r1s:
Remove redundant  type hint, which was caused by typo or smth like that.
Those type hints completely equals, so that does not make any sense.


CHB-0r1s:
@abetlen Hi, just a ping

CHB-0r1s:
@abetlen Hi, just a ping

PR #1973, opened 2025-03-17
Update README.md For llama.cpp HIPBLAS compile instructions

Scofflawvii:
[From llama.cpp repo -DGGML_HIP=on is now the required CMAKE flag for HIPBlas/ROCM.](https://github.com/ggml-org/llama.cpp/blob/master/docs/build.md#hip)

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,10 +175,10 @@ pip install llama-cpp-python \
 <details>
 <summary>hipBLAS (ROCm)</summary>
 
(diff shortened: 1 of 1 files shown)


agronholm:
Was [my PR](https://github.com/abetlen/llama-cpp-python/pull/1867) not good enough?

Scofflawvii:
My bad sorry about that.

PR #1980, opened 2025-03-22
sdist: include .ggml_build_number in tarball

### LINKED ISSUE #1979: Set GGML_BUILD_NUMBER to the correct version when building from pypi tarball
# Prerequisites

- [X] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [X] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [X] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [X] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

When I build the package from the source tarball uploaded to PyPI, I expect `PACKAGE_VERSION` in `ggml-version.cmake` file set to the same version as extracted from `git` when building directly from the git repo.

# Current Behavior

```
$ grep PACKAGE_VERSION result/lib/python3.12/site-packages/lib/cmake/ggml/ggml-version.cmake | grep '^set'
set(PACKAGE_VERSION "0.0.1"
```

This is probably because while the published tarball includes `.git`, the checkout done by `actions/checkout` in the publish workflow doesn't fetch full depth of the git repo, so the checkout turns out shallow.

This is what is seen during the build:

```
python3.12-llama-cpp-python> CMake Warning at vendor/llama.cpp/ggml/CMakeLists.txt:298 (message):
python3.12-llama-cpp-python>   GGML build version fixed at 1 likely due to a shallow clone.
```

I **think** this can be fixed by setting `fetch-depth` to `0` for the publish workflow. Of course, this may be not very optimal because then the resulting tarball will have to include complete git repo.

Perhaps an alternative fix is to calculate the version during the sdist build, then storing it inside the tarball as a file; finally, read the version from this file during the build, if it's present.

booxter:
With that, builds using tarball will still have the correct cmake
package version for ggml.

This assumes the file is read by llama.cpp build system, see:
https://github.com/ggml-org/llama.cpp/pull/12509

Fixes #1979

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ wheel.packages = ["llama_cpp"]
 cmake.verbose = true
 cmake.minimum-version = "3.21"
 minimum-version = "0.5.1"
-sdist.include = [".git", "vendor/llama.cpp/*"]
+sdist.include = [".ggml_build_number", "vendor/llama.cpp/*"]
 
 [tool.scikit-build.metadata.version]
 provider = "scikit_build_core.metadata.regex"
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -12,6 +12,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       with:
+        fetch-depth: 0
         submodules: "recursive"
 
     - name: Set up Python
@@ -41,6 +42,7 @@ jobs:
 
     - name: Build source distribution
       run: |
+        (cd vendor/llama.cpp && git rev-list --count HEAD) > .ggml_build_number
         python -m build --sdist
 
     - name: Publish distribution to PyPI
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -65,7 +65,7 @@ wheel.packages = ["llama_cpp"]
 cmake.verbose = true
 cmake.minimum-version = "3.21"
 minimum-version = "0.5.1"
-sdist.include = [".git", "vendor/llama.cpp/*"]
```
booxter, line 68: I'm not sure why `.git` was included before. I can build the package even when I remote the directory completely before the build.

booxter:
Will need to figure out how to test the change to github workflow actually works.

booxter:
I won't have time to get it to completion sadly. Feel free to take it over if anyone is interested.

PR #2018, opened 2025-05-12
feat: add Jinja2 loop controls to chat templates

handshape:
Hi there -- Cohere's Command models use `break` in their jinja templates. Adding the `jinja2.ext.loopcontrols` extesion to the jinja environment is sufficient to enable support.


handshape:
Now that devstral has dropped, and has even wackier stuff in their jinja templates, I wonder if there's a general form of this issue. Would it make sense to expose the jinja context, and let the consumer of the library decide what features and context objects they want to enable in jinja?

Issue #2026, opened 2025-06-03
llama_cpp/lib/libllama.so: undefined symbol: llama_kv_cache_view_init


opsec-ai:
# Prerequisites

**Just built with Python12 in fresh .venv**

Please answer the following questions for yourself before submitting an issue.

- [ x] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [x ] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [ x] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [ x] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

from llama_cpp import Llama
(success!)

# Current Behavior

from llama_cpp import Llama
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/k/Downloads/src/chatterbox/.venv/lib64/python3.12/site-packages/llama_cpp/__init__.py", line 1, in <module>
    from .llama_cpp import *
  File "/home/k/Downloads/src/chatterbox/.venv/lib64/python3.12/site-packages/llama_cpp/llama_cpp.py", line 1824, in <module>
    @ctypes_function(
     ^^^^^^^^^^^^^^^^
  File "/home/k/Downloads/src/chatterbox/.venv/lib64/python3.12/site-packages/llama_cpp/_ctypes_extensions.py", line 113, in decorator
    func = getattr(lib, name)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/ctypes/__init__.py", line 392, in __getattr__
    func = self.__getitem__(name)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/ctypes/__init__.py", line 397, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: /home/k/Downloads/src/chatterbox/.venv/lib64/python3.12/site-packages/llama_cpp/lib/libllama.so: undefined symbol: llama_kv_cache_view_init
>>>

# Environment and Context

$ sysinfo
[...]

PR #2030, opened 2025-06-13
Remove llama_kv_cache_view and deprecations were deleted on llama.cpp side too

serhii-nakon:
rel: https://github.com/ggml-org/llama.cpp/pull/13653
rel: https://github.com/abetlen/llama-cpp-python/issues/2026


serhii-nakon:
By some reason even with this fixes it crashes - if it possible review it. Maybe it because of another fix in my local env (related with template strftime_now)

serhii-nakon:
OK, I reverted back llama.cpp - it start works with fixes in this PR. Looks like here some more issues with new llama.cpp and llama-cpp-python compatibility.

PR #2049, opened 2025-08-06
Small updates to allow for `gpt-oss` generation

iamlemec:
Compatibility updates for `gpt-oss` generation:
- Seems like `seq_rm` doesn't accept `-1` anymore, so just use `seq_id = 0` for generation (is this fine for the server?)
- The `gpt-oss` template uses `strftime_now` in the template, so provide that function

Of course, this also needs a `llama.cpp` version bump (not included here) for the new model architecture.


abetlen:
Thank you @iamlemec !

Issue #2051, opened 2025-08-08
Can't compute multiple embeddings in a single call


jeberger:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest code. Development is very rapid so there are no tagged versions as of now.
- [x] I carefully followed the [README.md](https://github.com/abetlen/llama-cpp-python/blob/main/README.md).
- [x] I [searched using keywords relevant to my issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the [Discussions](https://github.com/abetlen/llama-cpp-python/discussions), and have a new bug or useful enhancement to share.

# Expected Behavior

Running this code:

```python
model = llama_cpp.Llama ("mxbai-embed-xsmall-v1-q8_0.gguf", embedding = True)
embeddings = model.embed (["Hello", "World"])
```

used to work in v0.3.14

# Current Behavior

The code raises an exception `RuntimeError: llama_decode returned -1`. The following messages are printed to the console:

```
init: invalid seq_id[3][0] = 1 >= 1
encode: failed to initialize batch
```

# Environment and Context

llama-cpp-python was compiled in CUDA mode

# Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

# Steps to Reproduce

```python
Python 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import llama_cpp
>>> model = llama_cpp.Llama ("../models/mxbai-embed-xsmall-v1-q8_0.gguf", embedding = True)
...
>>> embeddings = model.embed (["Hello", "World"])
decode: cannot decode batches with this context (calling encode() instead)
init: invalid seq_id[3][0] = 1 >= 1
encode: failed to initialize batch
llama_decode: failed to decode, ret = -1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../site-packages/llama_cpp/llama.py", line 1108, in embed
    decode_batch(s_batch)
  File ".../site-packages/llama_cpp/llama.py", line 1045, in decode_batch
    self._ctx.decode(self._batch)
  File ".../site-packages/llama_cpp/_internals.py", line 327, in decode
[...]

PR #2058, opened 2025-08-19
Fix multi-sequence embeddings

iamlemec:
Fixes multi-sequence (batch) embeddings by handling `n_seq_max` and `kv_unified` flags. See discussion in #2051.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -90,6 +90,8 @@ class Llama:
         yarn_orig_ctx: int = 0,
         logits_all: bool = False,
         embedding: bool = False,
+        n_seq_max: Optional[int] = None,
+        kv_unified: Optional[bool] = None,
         offload_kqv: bool = True,
         flash_attn: bool = False,
         op_offload: Optional[bool] = None,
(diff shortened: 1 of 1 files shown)


LimePencil:
@abetlen any updates yet?

freckletonj:
confirming this is still an issue

mlisovyi:
Shouldn't `n_seq_max` be also used in `Llama.embed()` ? One should add `or p_batch == n_seq_max` to the batch-evaluation condition in the loop [here](https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama.py#L1107). Otherwise one runs into a danger of collecting a batch that will consist of more sequences as the configured maximum (if individual inputs are short or the configured `n_seq_max` is small) and this will also lead to the same `llama_decode returned -1` error

mlisovyi:
Also, would it make sense to expose those parameters in `ModelSettings` with some meiningful defaults to allow setting them in the server run?

abetlen:
@iamlemec thank you, this should have been fixed in v0.3.23

PR #2077, opened 2025-10-12
support batch embeddings and zero-copy numpy returns

kavorite:
- Add n_seq_max parameter to `Llama` class to enable batch embeddings (defaults to 1 for backward compatibility)
- Add `return_numpy` support to convert between numpy arrays and lists with zero copies
- Update `normalize_embedding()` to keep numpy arrays as numpy arrays for zero-copy efficiency
- Update `test_embed_numpy` to use `n_seq_max=16` for batch embedding tests

Enables batch embedding support which was previously failing with llama_decode errors due to `n_seq_max=1` limitation. This also fixes a bug in a repo I was working on that consumes this functionality to mass index GitHub repos for semantic multivector search on the machine under my desk (luh mao).

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -25,6 +25,8 @@ from typing import (
     Deque,
     Callable,
     Dict,
+    Tuple,
+    overload,
 )
 from collections import deque
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


kavorite:
this is addressed in #2058, which is cleaner. I think perhaps I will keep this locally until that is merged; if it is reopened it will address the numpy piece only, and hopefully with less bloat due to spuriously altered formatting

PR #2100, opened 2025-12-10
Fix Errors in GBNF string constants

hi-cannon:
JAPANESE_GBNF and CHESS_GBNF are both copies of the `json.gbnf` from the llama.cpp repo. PR replaces this string with the proper `japanese.gbnf` and `chess.gbnf` examples from the repo

There is a significant error in the scientific notation portion of the JSON_GBNF string:
`([eE] [-+]? [0-9] [1-9]{0,15})`
This statement forbids trailing zeros, making `e+10`, `e+100` etc impossible to generate. PR swaps the last two character groups

SPACE_RULE is declared twice, PR removes duplicate

### DIFF
--- a/llama_cpp/llama_grammar.py
+++ b/llama_cpp/llama_grammar.py
@@ -104,31 +104,19 @@ ws ::= ([ \t\n]+)
 """
 
 CHESS_GBNF = r"""
-root   ::= object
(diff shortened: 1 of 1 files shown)


hi-cannon:
Doesn't look like these are actually used anywhere so it doesn't matter

PR #2109, opened 2026-01-04
feat: support Granite-Docling model

dhdaines:
Applies on top of #2108 which has the necessary changes to MTMD.

This adds chat formats to support https://huggingface.co/ggml-org/granite-docling-258M-GGUF and its ancestor, https://huggingface.co/ggml-org/SmolVLM-256M-Instruct-GGUF (and various other SmolVLM models)

In order to use Granite-Docling effectively for table structure, equation and layout extraction, it is necessary to enable special tokens in the chat completion output, so this adds a `special` flag to all of the chat completion functions which matches what `--special` does in `llama-cli` (this is enabled by default in `llama-mtmd-cli`)


dhdaines:
Ready for review!

PR #2133, opened 2026-03-04
feat: Qwen 3.5 GDN support with hybrid model fixes

r-dh:
## Summary

Adds Qwen 3.5 (Gated Delta Network) support, building on the work in #2132 by @codavidgarcia, with additional fixes needed to make hybrid GDN models work correctly:

- Update llama.cpp submodule and bindings for Qwen 3.5 (from #2132)
- Fix CMake build: set `BUILD_NUMBER` and `LLAMA_INSTALL_VERSION` for mtmd
- Return bool from `kv_cache_seq_rm` so callers can detect when partial memory removal fails
- Fix prefix-caching in `generate()`: GDN hybrid models reject partial memory removal via `llama_memory_seq_rm()`, but the return value was being ignored, causing `llama_decode returned -1` on subsequent calls. Now falls back to full prompt re-evaluation when partial removal is not supported.

The GDN prefix-caching fix affects all hybrid architecture models (not just Qwen 3.5), since any model mixing KV-cache attention with recurrent state will return `False` from `llama_memory_seq_rm()` for partial ranges.

## Test plan

- [x] Verified with Qwen3.5-4B-GGUF (Q4_K_M) at n_ctx=6144 on Apple Metal
- [x] Multi-turn inference (prefix-caching across calls) works correctly
- [x] Embeddings work with updated `llama_memory_clear` API
- [x] RAG and agentic tool-calling flows pass end-to-end

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -892,13 +892,23 @@ class Llama:
                 else:
                     break
             if longest_prefix > 0:
-                reset = False
-                tokens = tokens[longest_prefix:]
-                self.n_tokens = longest_prefix
-                if self.verbose:
+                # Try to trim the KV cache to prefix length.  Hybrid models
+                # (e.g. GDN) may not support partial removal — in that case we
+                # fall through to the full reset path below.
+                if self._ctx.kv_cache_seq_rm(-1, longest_prefix, -1):
+                    reset = False
+                    tokens = tokens[longest_prefix:]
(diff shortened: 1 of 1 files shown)


HoochDeveloper:
Thanks @r-dh this is much appreciated!

abetlen:
@r-dh thanks for the contribution, changes included in #2152

PR #2164, opened 2026-03-26
ci: refactor cpu wheel build workflow

Bing-su:
fixes #2163

Builds widely reusable wheels.

In fact, since llama-cpp-python uses ctypes, there is no need to specify a Python ABI. Therefore, tags like `py3-none-manylinux_2_34_x86_64.whl` can be used.

Please check the build success log at this link:
https://github.com/Bing-su/llama-cpp-python/actions/runs/23591453286

### DIFF
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -94,6 +94,15 @@ jobs:
           submodules: "recursive"
 
       - uses: astral-sh/setup-uv@v7
+        with:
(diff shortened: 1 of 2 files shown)


abetlen:
@Bing-su thank you for the contribution, these are some really good changes.

I'm going to likely break this up across a few different PRs to ensure it's easy to rollback in case anything goes wrong but I'll be sure to credit you in the CHANGELOG.

PR #2167, opened 2026-03-30
feat: add `reasoning_effort` to chat completions API

abetlen:
Adds OpenAI-compatible `reasoning_effort` support on chat completions and forwards it to chat handlers/templates as an optional keyword argument.


abetlen:
@dimitribarbot added that via #2168 unless you mean the ability to set this in the request?

Niels-BW:
> > @dimitribarbot added that via #2168 unless you mean the ability to set this in the request?
> 
> Yes, I meant to add that parameter to the `CreateChatCompletionRequest` and to the `create_chat_completion` function in `llama_cpp/llama.py`, just as you did for `reasoning_effort`.

This would be a very valuable addition for me as well

PR #2171, opened 2026-04-03
chore: bump version to 0.3.20

abetlen:


jwijffels:
Thanks! Works nicely with gemma-4-e2b-it-Q8_0 from https://huggingface.co/ggml-org/gemma-4-E2B-it-GGUF/tree/main when I tested it. Many thanks for the quick changes.

PR #2172, opened 2026-04-03
fix(ci): avoid duplicate py3-none release builds

abetlen:


abetlen:
Closing as stale/superseded. The relevant py3-none release-wheel fixes have since landed through later CI PRs, and the current release workflows successfully published v0.3.22. This branch now conflicts with main and includes older workflow state that would regress the CUDA matrix.

PR #2180, opened 2026-04-12
fix: prevent KV cache corruption on SWA/ISWA models + hot-path perf

avion23:
## Problem

Gemma-4 and any model with interleaved sliding window attention (ISWA) crashes on the second call to `create_chat_completion`:

```python
llm = Llama(model_path="gemma4-q4.gguf", n_gpu_layers=-1)

llm.create_chat_completion(messages=[{"role": "user", "content": "What is 2+2?"}])
# → OK

llm.create_chat_completion(messages=[{"role": "user", "content": "Write a hello world"}])
# → RuntimeError: error during generation: [end of text]
```

### Root cause

ISWA KV caches store position tracking in global maps (`g_iswa_pos_max` / `g_iswa_pos_min`) that are cleared by `llama_memory_clear()` but **not** by `llama_memory_seq_rm()`. The `generate()` method detects a prefix match between consecutive prompts (shared BOS token), calls `kv_cache_seq_rm()` to remove the divergent tail, sees it return `True`, and skips the full reset. Stale position maps cause batch allocator inconsistency → `llama_decode` returns -1.

Additionally, `reset()` was a no-op on KV cache state (only reset `n_tokens`), so calling `llm.reset()` between prompts had no effect.

A separate `tokens[:-1]` off-by-one in prefix matching silently broke prefix detection for all models.

### Fix

1. `reset()` calls `llama_memory_clear()` unconditionally — proper KV cache state reset
2. `generate()` sets `longest_prefix = 0` for recurrent/SWA models when the prefix doesn't cover all cached tokens, falling through to full reset
3. Fixed `tokens[:-1]` → `tokens` in prefix matching
4. `eval()` simplified: removed dead code paths, direct 2D logits assignment

## Performance

| Location | Before | After |
|---|---|---|
| `set_batch()` | Python loop: 512 tokens × 5 assignments | 5 numpy bulk writes |
| `_create_completion` detokenization | O(n²) per-token `detokenize(completion_tokens)` | O(1) incremental `token_to_piece()` |
| `_create_completion` logit_bias | `np.copy(scores)` every token | in-place modification |
| `_create_completion` top-k logprobs | `sorted()` on 128K vocab: O(V log V) | `np.argpartition`: O(V) |
| `eval()` logits | flattened copy + reshape | direct 2D view assignment |
| `token_to_piece()` | `create_string_buffer(32)` with trailing garbage | returns exact byte length |

### Benchmarks (TinyLlama-1.1B, M1 Pro, 10-50 runs)

| Test | Result |
|---|---|
| Generation throughput | ~930 tokens/sec |
| set_batch (200 token prompt) | ~61ms |
| logit_bias processing | ~54ms |
| top-k argpartition vs sorted (128K vocab) | **32x faster** (888ms vs 28.3s for 1000 runs) |

## Testing

- **ISWA**: Gemma-4-E2B-it (Q4_K_M, Q8_0) — multi-turn 6 rounds ✅
- **Ministral-3**: 3B — multi-turn 3+ rounds ✅; 8B, 14B — multi-turn 3+ rounds ✅ (requires n_ctx=2048)
- **Other**: Mistral-7B, Llama-3-8B, Phi-4, Gemma-2-9B, Qwen3-VL-8B, Mistral-Small-24B, Nomos-1, TinyLlama-1.1B ✅ (13 models total)

Functional verification: ISWA crash fix, reset(), prefix matching, logit_bias, set_batch, eval(), token_to_piece, argpartition correctness — all pass.

### DIFF
--- a/llama_cpp/_internals.py
+++ b/llama_cpp/_internals.py
@@ -182,9 +182,12 @@ class LlamaModel:
         return list(tokens[:n_tokens])
 
     def token_to_piece(self, token: int, special: bool = False) -> bytes:
-        buf = ctypes.create_string_buffer(32)
-        llama_cpp.llama_token_to_piece(self.vocab, token, buf, 32, 0, special)
-        return bytes(buf)
+        size = 32
+        buffer = (ctypes.c_char * size)()
+        n = llama_cpp.llama_token_to_piece(
+            self.vocab, llama_cpp.llama_token(token), buffer, size, 0, special
+        )
+        return bytes(buffer[:n])
 
     def detokenize(self, tokens: List[int], special: bool = False) -> bytes:
         output = b""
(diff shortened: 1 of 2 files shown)


avion23:
@abetlen I think this would improve the software. What do you need to have this merged?

PR #2181, opened 2026-04-13
fix: suppress stdout and stderr in Jupyter notebooks

### LINKED ISSUE #872: suppress_stdout_stderr failing in notebooks

Anai-Guo:
## Problem

`suppress_stdout_stderr` does not suppress output in Jupyter notebook environments (VS Code notebooks, JupyterLab, etc.) when loading models with `verbose=False`.

This happens because ipykernel replaces `sys.stdout`/`sys.stderr` with `OutStream` objects that hold their own copy of the original file descriptor in `_original_stdstream_copy`. The `os.dup2` redirect used by `suppress_stdout_stderr` only affects the standard fd (1/2), but ipykernel's `OutStream` writes through its own fd copy, bypassing the redirect entirely.

## Fix

Before redirecting, check if `sys.stdout`/`sys.stderr` have the `_original_stdstream_copy` attribute (set by ipykernel). If so, temporarily point it at the standard fd so it gets redirected along with everything else. On exit, restore the original value.

This is a no-op in non-notebook environments since `_original_stdstream_copy` won't exist.

## Related

- Fixes #872
- ipykernel upstream issue: https://github.com/ipython/ipykernel/issues/795
- Workaround by @slundberg in #872 comments


Anai-Guo:
Friendly ping @abetlen — small fix for suppress_stdout_stderr inside Jupyter. Anything I can adjust?

PR #2183, opened 2026-04-13
Add chat template for gemma models

C00kieFact0ry:
## Problem
Using the standard Llava15ChatHandler with multimodal Gemma models fail. 
Gemma models expect `<start_of_turn>user` / `<start_of_turn>model` / `<end_of_turn>` tokens instead. There is no vision-capable handler subclass for Gemma models, even though the library already follows the pattern of subclassing `Llava15ChatHandler` for other model families (e.g., `ObsidianChatHandler`, `MoondreamChatHandler`, `Llava16ChatHandler`) in `llama_chat_format.py`.

Although there is a `format_gemma` for text based Gemma models, this is bypassed by the  [chat handler](https://github.com/C00kieFact0ry/llama-cpp-python/blob/1b1a320de18c14c3915ba2df59eedb7c6e7cbe69/llama_cpp/llama.py#L2012)

## Solution
For multimodal models, we need to change the user/assistant roles and add the proper start and end tokens using a [GemmaChatHandler](https://github.com/C00kieFact0ry/llama-cpp-python/blob/fdda9bcb27c9456ae09db55282898ef5de307b6d/llama_cpp/llama_chat_format.py#L3232)

### DIFF
--- a/llama_cpp/llama_chat_format.py
+++ b/llama_cpp/llama_chat_format.py
@@ -3229,6 +3229,62 @@ class Llava15ChatHandler:
         )
 
 
+class GemmaChatHandler(Llava15ChatHandler):
+    """Chat handler for Gemma-based multimodal models (e.g., PaliGemma, MedGemma).
+
+    Gemma models use <start_of_turn>/<end_of_turn> control tokens instead of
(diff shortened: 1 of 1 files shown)


C00kieFact0ry:
@abetlen Please check at your convenience🙏.

PR #2334, opened 2026-07-09
Fix: handle deprecated 'embedding' kwarg with warning

### LINKED ISSUE #2210: Llama() silently accepts and discards `embedding` kwarg; .embed() then raises confusingly
# Prerequisites

- [x] I carefully followed the README.md.
- [x] I searched using keywords relevant to my issue to make sure that I am creating a new issue that is not already open (or closed).
- [x] I reviewed the Discussions, and have a new bug or useful enhancement to share.

# Expected Behavior

When constructing `Llama` with the older spelling `embedding=True` (singular — the parameter name in 0.2.x), one of two things should happen:

1. The kwarg is accepted as a deprecated alias of `embeddings` and a `DeprecationWarning` is emitted, OR
2. A `TypeError` is raised at construction time, surfacing the issue at the call site rather than swallowing it silently.

# Current Behavior

Neither happens. `embedding=True` is silently swallowed via `**kwargs`, `context_params.embeddings` stays at its default `False`, and the failure surfaces much later — deep inside `.embed()` — with a misleading error message that suggests the user *didn't* pass the flag, when in fact they did (just under the historical name).

```
RuntimeError: Llama model must be created with embeddings=True to call this method
```

This is especially painful for users integrating older libraries that haven't migrated to the new spelling yet — the error points at the wrong thing.

# Environment and Context

- Hardware: x86_64, NVIDIA GeForce RTX 4090
- OS: Windows 10 22H2
- Python 3.12.9
- llama-cpp-python 0.3.36 (CUDA 12.8 prebuilt wheel)

```
$ python --version
Python 3.12.9

$ pip show llama-cpp-python | findstr Version
Version: 0.3.36
```

The constructor's `**kwargs` swallows unknown keyword arguments with no warning, so a typo or stale parameter name produces a delayed, confusing failure rather than an immediate error.

# Steps to Reproduce

```python
from llama_cpp import Llama

# Pass the older `embedding` (singular) instead of `embeddings` (plural).
m = Llama(model_path="path/to/model.gguf", embedding=True)
m.embed("hello")
```

Result:
```
RuntimeError: Llama model must be created with embeddings=True to call this method
```

Even though `embedding=True` was passed at construction. The fix is to either accept `embedding` as a deprecated alias or to validate kwargs strictly.

# Failure Logs

```
  File "...\Lib\site-packages\llama_cpp\llama.py", line 1602, in embed
    raise RuntimeError(
RuntimeError: Llama model must be created with embeddings=True to call this method
```

Hit while integrating Tencent's HY-Motion text-to-motion model — the `hymotion` package's text encoder still uses the older `embedding=` spelling, so anyone running it against llama-cpp-python 0.3.x sees this confusing failure at first inference instead of at construction. Workaround in our case is a runtime monkey-patch that translates `embedding` → `embeddings` in `Llama.__init__`, but that doesn't help anyone else.

lxcxjxhx:
Fixes #2210

## Problem
When users create a model with \Llama(embedding=True)\, the \embedding\ parameter is silently discarded by \**kwargs\, causing a \RuntimeError\ when calling \embed()\ method later.

## Solution
- Renamed the parameter from \embedding\ to \embeddings\ (plural form) to match \self.context_params.embeddings\
- Added backward compatibility: if users still use the old \embedding\ parameter, a \DeprecationWarning\ is issued and the value is automatically assigned to \embeddings\

## Testing
Verified that:
1. \Llama(embeddings=True)\ works correctly
2. \Llama(embedding=True)\ still works but shows a deprecation warning
3. The \embed()\ method can be called successfully in both cases

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -211,17 +211,6 @@ class Llama:
             embedding = kwargs.pop('embedding')
 
         self._stack = contextlib.ExitStack()
-
-        # Handle deprecated 'embedding' kwarg (singular) as alias for 'embeddings' (plural)
-        if 'embedding' in kwargs:
-            warnings.warn(
-                "The 'embedding' parameter is deprecated. Use 'embeddings' instead. "
-                "Support for 'embedding' will be removed in a future version.",
-                DeprecationWarning,
-                stacklevel=2
-            )
-            embedding = kwargs.pop('embedding')
-
         set_verbose(verbose)
 
         if not Llama.__backend_initialized:
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -90,7 +90,7 @@ class Llama:
         yarn_beta_slow: float = 1.0,
         yarn_orig_ctx: int = 0,
         logits_all: bool = False,
-        embedding: bool = False,
+        embeddings: bool = False,
         offload_kqv: bool = True,
         flash_attn: bool = False,
         op_offload: Optional[bool] = None,
@@ -208,7 +208,7 @@ class Llama:
                 DeprecationWarning,
                 stacklevel=2
             )
-            embedding = kwargs.pop('embedding')
+            embeddings = kwargs.pop('embedding')
 
         self._stack = contextlib.ExitStack()
         set_verbose(verbose)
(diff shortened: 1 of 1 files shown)


lxcxjxhx:
Closing: no maintainer activity since 2026-07-09. Will reopen if needed.

Issue #2342, opened 2026-07-17
Llama.close() doesn't clean up chat_handler causing null pointer on second load of a vision model


WarmneoN:
In ComfyUI's ToriiGate Llama.cpp Vision Generate node, when `keep_model_alive` is set to `false`, the first generation works, but the second inference crashes.

Root cause: when unloading the model, `Llama.close()` does not clean up the `chat_handler`. On second load, the C++ layer is polluted and `mtmd_ctx` becomes a null pointer.

Temporary fix: in `nodes_api.py` `_drop_local_llama`, before `del llm`:

```python
ch = getattr(llm, "chat_handler", None)
if ch is not None and hasattr(ch, "_exit_stack"):
    try:
        ch._exit_stack.close()
    except Exception:
        pass
```

Environment: Python 3.13 · PyTorch 2.9.1+cu130 · ComfyUI v0.28.0 · llama-cpp-python 0.3.34

PR #2353, opened 2026-07-31
fix: clean up chat handler resources in Llama.close()

JingliangGao:
When a vision model's chat_handler (e.g. Llava15ChatHandler, Qwen2VLChatHandler) is passed to Llama(), its _exit_stack holds C-level resources (mtmd_ctx). Calling Llama.close() previously skipped this cleanup, so reloading a second model would crash with a null pointer because the old mtmd_ctx was still alive.

Now close() checks for _exit_stack or a close() method on chat_handler and calls it before closing the main ExitStack.

### DIFF
--- a/llama_cpp/llama.py
+++ b/llama_cpp/llama.py
@@ -2279,6 +2279,12 @@ class Llama:
 
     def close(self) -> None:
         """Explicitly free the model from memory."""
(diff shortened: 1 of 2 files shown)


JingliangGao:
@abetlen Could you please review this PR when you have a moment? It fixes the crash issue described in #2342. Thanks!

